PR #5, opened 2021-10-11
MRG, CI: Restore CIs

### LINKED ISSUE #2: [MAINT] Setting up CI with GitHub Actions
As suggested, I set up CI for the associated tests for the pg-backend.

Currently to run viz-tests for the pg-backend, just the parameter of the fixture `browse_backend` hast to be set to "pyqtgraph".

My first shot on this was to create `conftest.py` within this repository for pyqtgraph and to copy the test-files from mne to run them in the directory of this repo (I didn't find a way to pass parameters to pytest when it runs in mne-python to just run "pyqtgraph").
There might be a better solution. WDYT?

With this solution we could maybe temporarily remove "pyqtgraph" from `browse_backend` in MNE-Python in the current integration PR to make it pass the tests there and work on passing the tests from this repo here.

larsoner:
Let's see how close I got

Closes #2

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -15,8 +15,6 @@ jobs:
       matrix:
         include:
         - os: ubuntu-latest
-          env:
-            DISPLAY: ':99.0'
         - os: windows-latest
         # - os: macos-latest
     defaults:
@@ -26,6 +24,7 @@ jobs:
       MNE_LOGGING_LEVEL: 'warning'
       MKL_NUM_THREADS: '1'
       PYTHONUNBUFFERED: '1'
+      DISPLAY: ':99.0'
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -5,33 +5,46 @@ on:
       branches: main
 
 jobs:
-#  raw-viz-tests:
-#    runs-on: ${{ matrix.os }}
-#    strategy:
-#      fail-fast: false
-#      matrix:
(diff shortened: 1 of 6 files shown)


marsipu:
> @marsipu feel free to merge. Now only macOS is disabled, and that's because there is a seemingly legitimate failure:
Thanks a lot @larsoner for solving those issues and doing it that quickly.

> https://github.com/mne-tools/mne-qt-browser/runs/3859408679

I see, I know that issue. For some reason "fake-clicking" on QPushButtons and QCheckBoxes seems to be inconsistent between OS and maybe also there setup. In some cases it works flawless while in other cases it doesn't. I tried around a lot to get to the bottom, but in the end I figured that in those cases we can be very sure that a button works if we call the selected slot.

I will fix that in the PR and then MacOs should also work here, right?

.github/workflows/qt_viz_tests.yml
```diff
@@ -1,37 +1,78 @@
...
-#        shell: bash
+  pytest:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - os: ubuntu-latest
+        - os: windows-latest
+        # - os: macos-latest
```
marsipu, line 19:
```suggestion
        - os: macos-latest
```

marsipu:
Hm, there is now a [SyntaxError](https://github.com/mne-tools/mne-qt-browser/pull/5/checks?check_run_id=3860067594) for MacOs in `Download testing data` which I don't really understand:
The python-version (3.9.7) is high enough for f-strings.
@larsoner Is this something you encountered too during fixing the setup as another reason for disabling MacOS?

larsoner:
All better

agramfort:
perfect timing ;)

>

PR #15, opened 2021-11-03
FIX: Improve the command

larsoner:
This should allow us to revert https://github.com/conda-forge/staged-recipes/pull/16711/commits/34d48a52cc6d89cde4dbb9ab82d5d6028b15e7d4. Currently running the command will silently download the `sample` dataset, which is not good. It also has no `--help` option, so the default `entry_point` actually tried to display -- which is problematic for the conda-forge recipe.


marsipu:
Thank you @larsoner. This entry-point was more meant for testing than for normal use. I did not know that it is invoked by conda-forge.
Should we add the --help then in this stage?
A future feature is planned where calling mne-qt-browser (or mne browse_raw) where you can open the browser without any file first and then select the file from a dialog. Then the --help would indicate that behaviour

larsoner:
> Should we add the --help then in this stage?

This PR at least allows things to be tested with `mne-qt-browser --help` and doesn't ever download `sample` itself, which are both improvements. But I'm also fine with removing the command altogether if it's really just for debugging.

> A future feature is planned where calling mne-qt-browser (or mne browse_raw) where you can open the browser without any file first and then select the file from a dialog. Then the --help would indicate that behaviour

To me this should just be `mne browse_raw` where you have `MNE_BROWSE_BACKEND=pyqtgraph` in your config, no need for a separate command.

So I think overall I'm +1 on removing `mne-qt-browser` as a command. Let me know if you agree and I'll change this PR to do it.

marsipu:
Yes lets remove it.

larsoner:
Done @marsipu

marsipu:
Thank you @larsoner

PR #17, opened 2021-11-04
FIX: Use upstream/main for tools

larsoner:
1. Don't install `upstream/main` for now, use stable
2. Use `upstream/main` rather than `marsipu/pg_test`

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -57,13 +57,13 @@ jobs:
         run: |  # use MNE main to ensure test files are available
           set -e
           python -m pip install --upgrade pip
-          git clone -b ${MNE_BRANCH} --single-branch --depth=1 https://github.com/mne-tools/mne-python.git
+          git clone -b ${MNE_BRANCH} --single-branch --depth=1 https://github.com/mne-tools/mne-python.git ../mne-python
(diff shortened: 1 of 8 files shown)


larsoner:
@marsipu this:

1. implements the os/mne matrix, plus an old/compat 3.7 version and no opengl variants -- so 9 variants (bad) but each is only ~5 min (good) so overall a good balance of coverage versus CI burden
2. injects an `ignore` for the warning about not having OpenGL (hopefully temporary and we'll go with opt-in so this warning goes away)
3. revamps the tests and dev instructions so that things are simpler
4. runs the benchmarks on CIs in addition to mne-python unit tests

With this I think we can ensure that things work. Ideally we would have a nightly scheduled job that emailed us on failure. One option is to switch to Azure Pipelines, because I know how to configure such nightly emails there. It looks like [there might not be a good solution for GitHub Actions currently](https://github.community/t/no-email-notification-on-schedule-cron-job-failure/119638/10). But in the meantime I think this PR is a step in the right direction, so feel free to merge if you're happy!

larsoner:
@marsipu I'm going to merge this so that the opt-in behavior can be tested properly, but feel free to comment and I'm happy to open another PR.

marsipu:
Thanks a lot @larsoner for taking care of that

PR #24, opened 2021-11-08
FIX: Threading issues

### LINKED ISSUE #19: BUG: Threading issues
Running
```
import mne
mne.viz.set_browser_backend('pyqtgraph')  # TODO: MNE_BROWSE_ but set_browser (r)
fname = mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(fname)
for _ in range(10):
    raw.plot()
```
I get a bunch of plots that show up, some of which are correct:

Screenshot from 2021-11-04 13-20-16

And a bunch of which never finish loading:

Screenshot from 2021-11-04 13-20-24

I suspect that this is a threading problem.

marsipu:
#### Description
When multiple plots of the same raw-object are started sequentially, probably a race condition between those threads accessing the same data prevent some threads from finishing.
Since only a single thread is used at startup of the plot to maintain interactivity, capping the `maxThreadCount` to 1 solves the problem (for multiple instances all Loading-Threads are queued and run sequentially).

When exiting the plot before the thread finished this could lead to unexpected hanging or even crashes. This PR also adds a cleanup for the threads when the application closes.

Fixes #19

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3684,7 +3684,7 @@ class PyQtGraphBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 QSettings().setValue(qsetting, value)
             self._close(event)
 
-        if self.load_thread is not None:
+        if hasattr(self, 'load_thread') and self.load_thread is not None:
             if self.load_thread.isRunning():
                 wait_time = 10  # max. waiting time in seconds
                 logger.info('Waiting for Loading-Thread to finish... '
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2079,6 +2079,11 @@ class PyQtGraphBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         self.mne.overview_mode = 'channels'
         self.mne.zscore_rgba = None
 
+        # Limit Thread-Count to 1 to avoid race-condition e.g. when the same
+        # raw is plotted multiple times and multiple instances access the same
+        # data.
+        QThreadPool().globalInstance().setMaxThreadCount(1)
+
         # Load from QSettings if available
         for qparam in qsettings_params:
             default = qsettings_params[qparam]
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3698,7 +3698,8 @@ class PyQtGraphBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 self.load_thread.wait(int(wait_time * 1e3))
 
         # Remove self from browser_instances in globals
-        _browser_instances.remove(self)
+        if self in _browser_instances:
+            _browser_instances.remove(self)
 
         self.gotClosed.emit()
         # Make sure PyQtBrowser gets deleted after it was closed.
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -48,6 +48,8 @@ from mne.io.pick import (_DATA_CH_TYPES_ORDER_DEFAULT,
                          channel_indices_by_type, _DATA_CH_TYPES_SPLIT)
 from mne.utils import logger, sizeof_fmt, warn, get_config
 
+from . import _browser_instances
+
 try:
     from pytestqt.exceptions import capture_exceptions
 except ImportError:
@@ -2053,6 +2055,10 @@ class PyQtGraphBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
     def __init__(self, **kwargs):
         self.backend_name = 'pyqtgraph'
 
+        # Add to list to keep a reference and avoid premature
+        # garbage-collection.
+        _browser_instances.append(self)
+
         BrowserBase.__init__(self, **kwargs)
         QMainWindow.__init__(self)
 
@@ -3690,6 +3696,9 @@ class PyQtGraphBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 logger.info('Waiting for Loading-Thread to finish... '
                             f'(max. {wait_time} sec)')
                 self.load_thread.wait(int(wait_time * 1e3))
+
+        # Remove self from browser_instances in globals
+        _browser_instances.remove(self)
 
         self.gotClosed.emit()
         # Make sure PyQtBrowser gets deleted after it was closed.
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -17,9 +17,8 @@ from functools import partial
 from os.path import getsize
 
 import numpy as np
-from PyQt5.QtCore import (QEvent, Qt, pyqtSignal, QRunnable,
-                          QObject, QThreadPool, QRectF, QLineF, QPoint,
-                          QSettings)
+from PyQt5.QtCore import (QEvent, QThread, Qt, pyqtSignal, QRectF, QLineF,
+                          QPoint, QSettings)
 from PyQt5.QtGui import (QFont, QIcon, QPixmap, QTransform,
                          QMouseEvent, QImage, QPainter, QPainterPath)
 from PyQt5.QtTest import QTest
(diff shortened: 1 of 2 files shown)


larsoner:
@marsipu I am testing by building the MNE-Python doc while on this `fix_threading` branch of `mne-qt-browser` and the `larsoner:pg_scraper` branch of MNE-Python, then running in MNE-Python's `doc` dir:
```
PATTERN=overview make html_dev-pattern
```
It just locks up.

So does:
```
$ MNE_BROWSE_BACKEND=pyqtgraph python
>>> import mne
>>> raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif')
>>> raw.plot()
```
So I don't think this fix is tenable.

marsipu:
@larsoner It's too late for me to check the docs now, I will try it in the next days.

But I couldn't reproduce the behaviour with the python console. I see this:
```
>>> raw.plot()
<mne_qt_browser._pg_figure.PyQtGraphBrowser object at 0x0000014D7E063040>
>>> Waiting for loading threads to finish...
...done!
```
...and I get the command-line back when I start typing

The messages indicates that the QApplication seems to send `aboutToQuit` but it is somehow kept alive by the Python-Console when the command-line is given back to the user.

larsoner:
I think it's probably the same issue we had before in https://github.com/mne-tools/mne-python/pull/9687#issuecomment-940215813

One way to replicate would be to spin up an Ubuntu VM (e.g., from [here](https://www.osboxes.org/ubuntu/)) on your Windows machine, and `pip install --user mne-qt-browser`, you should be able to replicate fairly quickly hopefully

marsipu:
I set up a spare laptop with Ubuntu 20.04 now.
And I still couldn't reproduce, that this locks up in the python-Terminal:
```
>>> import mne
>>> mne.viz.set_browser_backend()
>>> raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif')
>>> raw.plot()
```
neither this from https://github.com/mne-tools/mne-python/pull/9687#issuecomment-940215813:

```
python -ic "import matplotlib.pyplot as plt; plt.ion(); import mne; mne.viz.set_browser_backend('pyqtgraph'); mne.io.read_raw(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif').plot()"

```

But I could reproduce that the for the scraper PyQtGraphBrowser stays dark and blocking when `precompute=True`.
Unfortunately I still can't find a specific reason for that, maybe the sphinx-gallery environment just doesn't allow separate threads or something similar?

On the other hand setting `QThreadPool().maxThreadCount(1)` seems to fix the problem you originally mentioned in #19 (with the multiple function calls) while the lines from above still seem to work (at least with my Linux-Laptop).
WDYT @larsoner?

larsoner:
I am away from my Linux machine so can't test, but wasn't `QThreadPool().maxThreadCount(1)` in your original implementation and problematic? I seem to remember you having this or something a lot like it and us having to remove it.

> And I still couldn't reproduce, that this locks up in the python-Terminal:

Can you output your `mne sys_info` on your laptop? FWIW this is my `~/.config/matplotlib/matplotlibrc`:
```
backend : Qt5Agg
interactive : True
pdf.fonttype : 42
figure.max_open_warning : 100
```
But I doubt it matters since it should default to Qt5Agg and also the code I wrote above enables interactive mode. I would also try `pip install --upgrade pyqt5` on your laptop to see if it's a newer Qt issue for example.

marsipu:
> I am away from my Linux machine so I can't test, but wasn't QThreadPool().maxThreadCount(1) in your original implementation and problematic? I seem to remember you having this or something a lot like it and us having to remove it.

Yes, we had the discussion you pointed out. I assumed that limiting the thread-count to 1 would block the PyQtBrowser if matplotlib was somehow using a thread from the global ThreadPool. I am not sure anymore if that correlation is correct.

larsoner:
From `sys_info` one potential relevant difference is I am using latest PyQt5 (5.15.3 I think?). Feel free to test with that locally. I'm away from my Linux machine until next week but I can test then

marsipu:
I changed to a lower API-level with using QThreads directly now.
What I discovered when playing around with the multiple-plot-problem in IPython (with `%gui qt5`) or the python terminal:
```python
import mne
raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif')
for i in range(10):
    raw.plot(title=str(i))
```
This problem also appears with QThreads. But I think the reason for the empty plots is not related to the threads, but to some of the plots being garbage-collected since their object is not used anymore.
If the code is a little bit altered, all plots work as expected.
```python
import mne
browsers = dict()
raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif')
for i in range(10):
    browsers[i] = raw.plot(title=str(i))
```
@larsoner Can you reproduce?

larsoner:
> This problem also appears with QThreads. But I think the reason for the empty plots is not related to the threads, but to some of the plots being garbage-collected since their object is not used anymore.

This could make sense. Most GUI libraries (PyVista and matplotilb at least) keep their own private or public list of instances to prevent this. I think you could (and should) probably do the same thing. One option is to make sure that when the plotter is closed, it `.pop`s itself from this global list so that everything can be GC'ed. Want to try implementing this to see if it fixes the problem? Maybe you could use the `aboutToExit` event or whatever you were hooking into. We "just" need to make sure all types of closing (clicking the X in the upper right, code-based exits, ...?) actually hit this code path so we don't end up accidentally keeping references around.

larsoner:
> I changed to a lower API-level with using QThreads directly now.

This sounds better in principle than relying on some global set of threads / a threadpool to do work for you, assuming it works!

marsipu:
> One option is to make sure that when the plotter is closed, it .pops itself from this global list so that everything can be GC'ed. Want to try implementing this to see if it fixes the problem? 

Which problem do you mean here? I would think that the problem of the multiple plot call is a user problem since you would expect objects you don't keep a reference to to be GCed at some point, right?

marsipu:
I just tried it with your scraper, it seems to work now:
grafik

larsoner:
> is a user problem since you would expect objects you don't keep a reference to to be GCed at some point, right?

Not with GUI elements typically, no. It's not how matplotlib (and thus `raw.plot` for the last decade) or PyVista work at least. GUI frameworks are expected to keep objects around automagically until the window is closed (for better or worse).

larsoner:
... or to put it differently, code like this has worked for a long time on matplotlib:
```
raw.plot()
raw.copy().filter(1, None).plot()
```
and on your backend it will be up to the whims of `gc` whether or not this works (for how long, or if it does at all) because, unlike matplotlib, your backend doesn't currently keep references to the GUI elements around. Forcing users to do this:
```
fig1 = raw.plot()
fig2 = raw.copy().filter(1, None).plot()
```
to avoid GC seems like a step backward for them (in addition to being backward incompatible), even though I agree it would be nice not to have to keep track of the GUI elements ourselves to prevent GC.

larsoner:
> I don't know if Qt got this "global object" list but I think the problem is also that there seem to happen garbage-collection even when the Window is still shown in some cases.

You don't need any Qt magic. "Just" create a Python list in your module namespace somewhere, add the `QWindow`s to it as they are created, and make each QWindow `pop` itself from the list when it gets closed.

This is more or less what PyVista is doing here:

https://github.com/pyvista/pyvista/blob/29aad0eb6e89e8dad0b0f9e6d82b5c7a36dd1b49/pyvista/plotting/plotting.py#L5023

marsipu:
@larsoner Thanks, that was much simpler than expected;)

larsoner:
Okay I managed to SSH into my Linux machine and I first confirmed that the old `fix_threading` branch I was on *did* have some hanging problem. Now that I'm using your latest commit, I don't have any hanging problem, even with two windows open (and just plan `raw.plot()` calls, no assignment)! I'll review -- is this good to go from your end if it's all good?

larsoner:
... I see a lot of red CIs. Let me know when I should look (i.e., once CIs are green, or you just tell me to ignore them)

marsipu:
I'll fix it

marsipu:
If everything passes I think this PR would be good to go now.

mne_qt_browser/_pg_figure.py
```diff
@@ -3690,7 +3689,21 @@ def closeEvent(self, event):
                 value = getattr(self.mne, qsetting)
                 QSettings().setValue(qsetting, value)
             self._close(event)
+
+        if hasattr(self, 'load_thread') and self.load_thread is not None:
```
larsoner, line 3693: I prefer this more compact pattern but it doesn't really matter here
```suggestion
        if getattr(self, 'load_thread', None) is not None:
```

larsoner (approved):

larsoner:
Thanks @marsipu !!!

marsipu:
Thanks @larsoner for helping me getting this done

PR #25, opened 2021-11-12
FIX: bug with parameter "start"

marsipu:
This fixes a bug where the data is not correctly showed when the parameter `start` is set to another value than `0`.


larsoner (approved):
LGTM feel free to merge if you're happy @marsipu

Issue #26, opened 2021-11-15
documentation of "zscore" and "channels" Overview-mode


sappelhoff:
Is there a documentation of what the "zscore" and "channels" overview modes are supposed to do?

I am talking about this panel (currently in zscore mode):

![image](https://user-images.githubusercontent.com/9084751/141773105-c7eeb508-5c0c-4832-98fe-001f78e52b5e.png)

I couldn't find any in https://mne.tools/dev/generated/mne.io.Raw.html?highlight=raw#mne.io.Raw.plot

marsipu:
This Overview-Bar might be subject to further change anyway (making channels-mode obsolete since its lack of information) as
@larsoner and @drammock discussed in #20

sappelhoff:
Okay, found it, thanks :-)

Yes, a "help" button would be better (more easily findable). An alternative to that might be to either:

1. let the hover-box appear much faster, and let it linger for a bit after the cursor moves away
2. force-open the hover-box when a user opens the "select box", that is, when they click on the dropdown menu to select either channels or zscore

marsipu:
I fixed this in #29.
I went for your first suggestion since I think that the tool tip is a bit quicker than an extra dialog for the help which has to be clicked away then. And just hovering lets you see what you might change before you change it.

PR #29, opened 2021-11-18
FIX: Show Tooltip immediately

marsipu:
This shows the tooltip of the Overview-ComboBox immediately as suggested by @sappelhoff in #26.


sappelhoff (approved):
I think that's a nice improvement, thanks!

PR #31, opened 2021-11-30
[MAINT] bump `__version__` to dev

sappelhoff:
I just installed mne-qt-browser at a specific commit **after** 0.1.6 (06e586a44fc7934b14ad7c0cde89f527bafbd879) to get an unreleased fix ... yet `mne sys_info` still showed me `0.1.6` as the version, which is a bit misleading.

I assumed you are following a "semver"-style of versioning for the proposed changes below.

### DIFF
--- a/mne_qt_browser/_version.py
+++ b/mne_qt_browser/_version.py
@@ -1,1 +1,2 @@
-__version__ = '0.1.6'
+"""The version number."""
+__version__ = '0.1.7.dev0'


marsipu:
Thank you @sappelhoff. So when I bump to a new version on PyPi and conda-forge, I need to:

1. bump `__version__` to the stable version
2. Upload to PyPi/merge the conda-forge PR
3. Then bump `__version__` to dev-version

Right?

sappelhoff:
>  need to:
>
>    bump __version__ to the stable version
>    Upload to PyPi/merge the conda-forge PR
>    Then bump __version__ to dev-version
>
>Right?

exactly, bumping to the next dev version should be the first commit after a release :+1:

larsoner:
Agreed, this is what we do in the packages I maintain and it works well. Thanks @sappelhoff !

PR #32, opened 2021-11-30
FIX: Accept all matplotlib color-specifiers

marsipu:
## Description
This PR adds a wrapper for pyqtgraph's `mkColor` which makes PyQtGraphBrowser accept [all available matplotlib color-specifiers](https://matplotlib.org/stable/tutorials/colors/colors.html) (especially float-tuples with values from 0-1 where pyqtgraph would need values from 0-255)

## References
Closes mne-tools/mne-python#10059

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -46,7 +46,7 @@ from mne.viz.utils import _simplify_float, _merge_annotations
 from mne.annotations import _sync_onset
 from mne.io.pick import (_DATA_CH_TYPES_ORDER_DEFAULT,
                          channel_indices_by_type, _DATA_CH_TYPES_SPLIT)
-from mne.utils import logger, sizeof_fmt, warn, get_config
+from mne.utils import _to_rgb, logger, sizeof_fmt, warn, get_config
 
 from . import _browser_instances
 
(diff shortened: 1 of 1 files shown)


mne_qt_browser/_pg_figure.py
```diff
@@ -69,6 +69,29 @@ def _get_std_icon(icon_name):
...
+    if (isinstance(color_spec, tuple) and
+            all([i <= 1 for i in color_spec])):
+        color_spec = tuple([int(i * 255) for i in color_spec])
+
+    try:
+        color = mkColor(color_spec)
+    except ValueError:
+        logger.critical(f'{color_spec} is not a valid matplotlib '
+                        f'color-specifier!')
+        return mkColor((0, 0, 0, 128))
```
larsoner, line 90: I think this should emit an error like
```
raise ValueError(...) from None
```

marsipu: Ok, so the plot will fail then

larsoner: Yeah I think if you give an invalid matplotilb color you should get an error and the plot should fail

mne_qt_browser/_pg_figure.py
```diff
@@ -69,6 +69,28 @@ def _get_std_icon(icon_name):
...
+    # Convert tuples of floats from 0-1 to 0-255 for pyqtgraph
+    if (isinstance(color_spec, tuple) and
+            all([i <= 1 for i in color_spec])):
+        color_spec = tuple([int(i * 255) for i in color_spec])
+
+    try:
+        color = mkColor(color_spec)
+    except ValueError:
+        raise ValueError(f'"{color_spec}" is not a valid matplotlib '
+                         f'color-specifier!')
```
larsoner, line 89:
```suggestion
                         f'color-specifier!') from None
```
because it's cleaner:
```
>>> try:
...     raise ValueError
... except Exception:
...     raise RuntimeError
... 
  File "<stdin>", line 2, in <module>
ValueError

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
RuntimeError
>>> try:
...     raise ValueError
... except Exception:
...     raise RuntimeError from None
... 
  File "<stdin>", line 4, in <module>
RuntimeError
```

larsoner (approved):
LGTM, will merge once CIs come back happy. Thanks @marsipu !

sappelhoff:
Thanks!

PR #42, opened 2022-02-09
add upload workflow

marsipu:
This adds an upload-workflow similar to [here](https://www.caktusgroup.com/blog/2021/02/11/automating-pypi-releases/).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,24 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [ created ]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: |
+          python setup.py sdist bdist_wheel
+          twine upload dist/*
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,4 +21,5 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
           python setup.py sdist bdist_wheel
+          twine check dist/*
           twine upload dist/*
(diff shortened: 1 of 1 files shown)


hoechenberger:
FYI I'm using using something similar for `openneuro-py`:
https://github.com/hoechenberger/openneuro-py/blob/main/.github/workflows/python-publish.yml

I'm doing once additonal things before uploading, which is running a `twine check`

Also I'm getting the username from `${{ secrets.PYPI_USERNAME }}`, i.e. it's stored in the GH secrets

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: |
+          python setup.py sdist bdist_wheel
+          twine upload dist/*
```
hoechenberger, line 24:
```suggestion
          twine check dist/*
          twine upload dist/*

```

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: __token__
```
hoechenberger, line 20: I think you should store it in the GH secrets and retrieve it via

```suggestion
          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```

larsoner: I wouldn't worry about exposing that you use a `__token__` for the upload. I don't think it really helps people hack anything (if anything, I'd say it discourages it since these are meant to be harder than user passwords to crack)

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
```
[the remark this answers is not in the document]
marsipu: Not common to upgrade pip? What is the question?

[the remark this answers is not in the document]
marsipu: Ah, you mean for L.17?

hoechenberger (approved):
Looking good!

PR #43, opened 2022-02-10
Update getting started guide & add screenshot

hoechenberger:
I tried to make it easier to get started by providing a single script (not split across several chunks) that users can just copy & paste. Using the `sample` dataset now so it it will work everywhere out of the box.

Also added a screenshot based on the branch with the new icons.

*Careful:* The embedded screenshot will break the rendering on PyPI because it's embedded into the readme via a relative link. Once this PR has been merged, we need to update the readme to include an absolute URL.

View the "rendered" readme here:
https://github.com/mne-tools/mne-qt-browser/blob/a729323786050575df84bdd9471d4e2857d328ce/README.md

WDYT?

cc @agramfort


README.md
```diff
@@ -23,46 +25,60 @@ or
...
-```
 
-Then load and plot your Raw-data, e.g. by using:
+If the plot is not showing, search for solutions in the
+[troubleshooting](#troubleshooting) section below.
+
+This will use the `mne-qt-browser` for the current Python session. If you
+want to make this change permanent so you don't have to use the
+`set_browser_backend()` anymore, run the following line to modify your MNE
+configuration file:
```
hoechenberger, line 52:
```suggestion
This will use the `mne-qt-browser` for the current Python session. If you
want to make this change permanent so you don't have to use the
`set_browser_backend()` each time after restarting Python, run the following
line to modify your MNE configuration file:
```

larsoner:
In MNE-Python we only support the current release (+dev), I would do the same for mne-qt-browser. So that means support 0.24+dev until 1.0 is released, then start supporting 1.0+dev here

marsipu:
I think this one is save to merged without #48 for passing the CIs (changes here are not relevant for tests).
Thank you @hoechenberger

PR #48, opened 2022-02-10
MNT: Downgrade pytest to pass tests with maint/0.24

marsipu:
This PR avoids all CIs to fail for the branch `maint/0.24` from mne-python by setting `pytest==6.2.5`.
The cause is a `DeprecationError` which is fixed in mne-tools/mne-python#10120.


marsipu:
@larsoner I can't get it to work, can you help?
I didn't know which line was the relevant one from the [link you send me](https://github.com/mne-tools/mne-python/edit/main/mne/conftest.py#L130-L136) and pytest doesn't seem to accept any string I try to set to filter "pytest.PytestRemovedIn8Warning: Passing None has been deprecated."

The other solution I had before you gave me the tip from above was just downgrading to the previous pytest-version. This works but is also only a temporary solution

marsipu:
No problem:thumbsup:, I'll do that

PR #40, opened 2022-01-30
ENH: Epochs support

marsipu:
## Description
This is work in progress for epochs support.

- [x] adapt pyqtgraph-backend to IO for Epochs from mpl-backend
- [x] handle visualization
- [x] handle epochs-specific interaction (selecting bad epochs, etc.)
- [x] let pyqtgraph-backend pass `mne.viz.tests.test_epochs.py`

This PR depends on a [correspondent PR](https://github.com/mne-tools/mne-python/pull/10297) in mne-python.

<details>
<summary>Code I currently use for testing</summary>

```python
import mne
import os
import numpy as np

from mne.viz import use_browser_backend

# Load Raw
sample_data_folder = mne.datasets.sample.data_path()
raw_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                        'sample_audvis_raw.fif')
raw = mne.io.read_raw(raw_file)

# Load Events
events_path = os.path.join(sample_data_folder, 'MEG', 'sample',
                           'sample_audvis_raw-eve.fif')
events = mne.read_events(events_path)

# Add test-annotations
onsets = np.arange(2, 8, 2) + raw.first_time
durations = np.repeat(1, len(onsets))
descriptions = ['Test1', 'Test2', 'Test3']
for onset, duration, description in zip(onsets, durations, descriptions):
    raw.annotations.append(onset, duration, description)

epo = mne.Epochs(raw, events)

epoch_col1 = np.asarray(['b'] * len(epo.ch_names))
epoch_col1[::2] = 'r'
epoch_col2 = np.asarray(['r'] * len(epo.ch_names))
epoch_col2[::2] = 'b'
epoch_col3 = np.asarray(['g'] * len(epo.ch_names))
epoch_col3[::2] = 'b'
epoch_colors = np.asarray([['b'] * len(epo.ch_names) for _ in
                           range(len(epo.events))])
epoch_colors[::3] = epoch_col1
epoch_colors[1::3] = epoch_col2
epoch_colors[2::3] = epoch_col3
epoch_colors = epoch_colors.tolist()

with use_browser_backend('pyqtgraph'):
    epo.plot(block=True, n_epochs=10, events=events, epoch_colors=epoch_colors,
             precompute=False)
```
</details>

**REMINDER**: CI testing with the PR branch from mne-tools/mne-python#10297 has to be reverted before merging!

PR #54, opened 2022-02-13
ENH: ICA support

marsipu:
This implements support for plotting ICA with `ICA.plot_sources(raw)` and when #40 is merged also `ICA.plot_sources(epochs)`.

grafik

#### Additional information
Depends on mne-tools/mne-python#10330

**REMINDER**: CI testing with the PR branch from mne-tools/mne-python#10330 has to be reverted before merging!


marsipu:
The last CIs are bound to fail, because https://github.com/mne-tools/mne-python/pull/10330 is not merged yet. But probably merging on mne-python's side first preferable

larsoner:
Green! Should we merge @marsipu ?

PR #55, opened 2022-02-16
BUG: Fix offbound overview-drag

marsipu:

This fixes the following minor bugs:
- [x] Dragging in Overview-Bar off bounds let to an error and unexpected behaviour
- [x] Set background-color white as default (relevant for Epochs.plot() without parameter "bgcolor")

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -995,32 +995,40 @@ class OverviewBar(QGraphicsView):
         x, y = self._mapToData(pos)
 
         # Set X
-        # Move click position to middle of view range
-        if self.mne.is_epochs:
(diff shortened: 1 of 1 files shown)


marsipu:
@larsoner @hoechenberger With e.g. a touchpad you can currently input vertical scrolling with steps smaller than the common scrollwheel-input. This results in smooth vertical scrolling. Would you rather keep this or do you think it is not necessary or even contraintuitive?

https://user-images.githubusercontent.com/46245704/154279087-a2b72536-4be2-4869-869d-8ff40441ee34.mp4

larsoner:
I think smooth is actually what most people would expect from the touchpad

hoechenberger:
Yes, smooth, anything else would feel wrong

PR #57, opened 2022-02-17
MNT: Adaptions related to mne-tools/mne-python#9960

marsipu:
This implements changes related to mne-tools/mne-python#9960.


larsoner:
Thanks @marsipu !

PR #59, opened 2022-02-17
FIX: Fixes

larsoner:
1. `orig_format` doesn't exist for Epochs/ICA, assume float32
2. `lasso_event` is not a real matplotlib event, let's just make our own `callbacks` and use them (this is in https://github.com/mne-tools/mne-python/pull/9960 but has a backward compat shim here)
3. When I did `group_by='selection'`, I actually had a matplotlib Qt window show up even though `show=False` was used in `plot_sensors`. A cleaner approach I think is just to explicitly use an Agg figure


marsipu:
I drafted the release but will wait until this is merged

marsipu:
Sure I can wait 👍🏻

larsoner:
Have to have this in `main` for the `[circle full]` run to pick it up.

PR #65, opened 2022-03-04
ENH: Add splash

larsoner:
Companion of https://github.com/mne-tools/mne-python/pull/10412

https://user-images.githubusercontent.com/2365790/156797553-1d9a8fab-a9cc-491f-b8f1-d9ca44eea823.mp4

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -42,6 +42,7 @@ from pyqtgraph import (AxisItem, GraphicsView, InfLineLabel, InfiniteLine,
                        mkPen, setConfigOption, mkColor)
 from scipy.stats import zscore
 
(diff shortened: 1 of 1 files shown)


larsoner:
We might want to do some sort of triaging based on the `show` value. Also it's better to not use `.finish` and instead wait until the `.show` completes. But this is at least a minimal POC

larsoner:
Okay this works well on macOS for me. We can merge once https://github.com/mne-tools/mne-python/pull/10412 is in, since this requires that private API to be available

larsoner:
@marsipu this is ready for review/merge from my end. It should work with latest MNE-Python `main`

marsipu (approved):

marsipu:
Thank you @larsoner

PR #66, opened 2022-03-04
FIX: Darkness

larsoner:
Before:
Screen Shot 2022-03-04 at 2 00 45 PM

After:
Screen Shot 2022-03-04 at 2 01 00 PM

PR #67, opened 2022-03-05
ENH: Invert colors

larsoner:
Just a little proof of concept about detecting dark mode and inverting all color lightness values using HSL or CIELAB to fit in better in dark mode. I'm not 100% sure we want this, but maybe we want some variant of it after https://github.com/mne-tools/mne-qt-browser/pull/66. CIELAB here requires `skimage` but there might be some pure-Python libs that have it. (Or maybe we can vendor skimage's conversion, especially since they require a 2D array which is a little bit annoying).

| `main` | ~~This PR, HSL~~ (defunct) | This PR, CIELAB |
| -- | -- | -- |
| Screen Shot 2022-03-05 at 5 02 27 PM | Screen Shot 2022-03-05 at 5 02 08 PM | Screen Shot 2022-03-05 at 5 16 07 PM |

To me CIELAB is much more legible, which I guess is to be expected.

cc @hoechenberger since I know you love dark mode and aesthetics :)


agramfort (approved):

larsoner:
No bikeshedding on HSL versus CIELAB? I'm disappointed in you @agramfort !

larsoner:
FYI, these two approaches are apparently [what Chrome uses to make dark mode on non-dark-mode websites](https://www.imore.com/google-chromes-slick-new-dark-mode-making-its-way-mac), so they at least are not totally unheard of. And they say:

> CIELAB is the more advanced of the two models, meaning it should produce more accurate and visually pleasing results. That being the case, all three of the "selective inversion" options are based on CIELAB

So I guess we should figure out how to use CIELAB. [colorspacious](https://pypi.org/project/colorspacious/) is small, on conda-forge, and by a MPL person, so that seems easier to add as a requirement than `skimage`.

API-wise, at the MNE-Python end we should maybe have a `theme` arg like [Brain](https://mne.tools/stable/generated/mne.viz.Brain.html#mne-viz-brain). Then we just have to decide if we want independent control of the trace viewport (light or dark mode, regardless of whether or not the other GUI elements are light or dark mode). Maybe YAGNI for now at least?

hoechenberger:
Great improvement!

larsoner:
Okay @marsipu I think we're ready to go here!

marsipu:
Nice, thank you @larsoner
Should we do a new release now?

larsoner:
I think so!

PR #87, opened 2022-04-01
ENH: Move overview selector to toolbar

### LINKED ISSUE #86: ENH: Move "overview mode" selector to toolbar
I understand the motivation for putting the "overview mode" selector at the bottom, because it places it closer to the overview bar. However, it creates a big waste of screen real estate, e.g. [here](https://mne.tools/dev/auto_tutorials/intro/15_inplace.html#sphx-glr-auto-tutorials-intro-15-inplace-py):

I propose to move it up into the toolbar. It's used rarely enough that I think the user experience degradation from having to move the mouse farther is acceptable given the screen space gains for showing traces.

cross-ref https://github.com/mne-tools/mne-python/issues/10481

Also, since the x-axis label is *always* either "time (sec)" or "epoch", perhaps it makes sense to move the label to the left of the numbers (aligned with the channel labels) or even just remove it altogether. I don't think people will be confused about what is being shown, and it gains us more vertical screen real estate.

Thoughts on these two ideas @marsipu @agramfort @hoechenberger ?

larsoner:
The only missing thing I think is to make a proper icon. @hoechenberger did you help with that / can you make one here and push?

This also gives an easy way to remove the bar because you can just deselect:

https://user-images.githubusercontent.com/2365790/161298432-0a6d1089-07b1-40d3-95ea-af9d920a33fa.mp4

Closes #86

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -27,8 +27,8 @@ from PyQt5.QtTest import QTest
 from PyQt5.QtWidgets import (QAction, QColorDialog, QComboBox, QDialog,
                              QDockWidget, QDoubleSpinBox, QFormLayout,
                              QGridLayout, QHBoxLayout, QInputDialog,
-                             QLabel, QMainWindow, QMessageBox,
-                             QPushButton, QScrollBar, QToolTip, QWidget,
+                             QLabel, QMainWindow, QMessageBox, QToolButton,
+                             QPushButton, QScrollBar, QWidget, QMenu,
                              QStyleOptionSlider, QStyle,
                              QApplication, QGraphicsView, QProgressBar,
                              QVBoxLayout, QLineEdit, QCheckBox, QScrollArea,
@@ -2565,19 +2565,6 @@ class LoadThread(QThread):
         del self.browser
 
 
-class _FastToolTipComboBox(QComboBox):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.setMouseTracking(True)
-
-    def setToolTip(self, tooltip):
-        self.tooltip = tooltip
-
-    def enterEvent(self, event):
-        QToolTip.showText(event.globalPos(), self.tooltip)
-        super().enterEvent(event)
-
-
 class _PGMetaClass(type(BrowserBase), type(QMainWindow)):
     """Class is necessary to prevent a metaclass conflict.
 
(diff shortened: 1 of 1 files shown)


larsoner:
... also I think we should get rid of the "fullscreen" button, modern UIs should provide this functionality I think.

cbrnr:
Wait, can you select more than one mode? +1 for removing the fullscreen button.

larsoner:
No, you cannot select more than one.

Also, I think the toolbar is more standard and looks better without labels:

Screenshot from 2022-04-01 12-08-56

The hover still shows you the full name, and I think the icons are clear enough that once you've done it once or twice you know what the buttons do.

larsoner:
(I can push a commit for this if others agree it's better)

larsoner:
Here is the equivalent text-beside-icon mode:
Screenshot from 2022-04-01 12-15-35

larsoner:
Okay I found a decent icon from [here](https://fonts.google.com/icons) called "view quilt":

Screenshot from 2022-04-01 12-22-34

Ready for review/merge from my end!

cbrnr:
I prefer no text, but this could also be a user-configurable setting.

If only one option is selectable, you should use radio buttons. Not sure how to deal with disabling the bar then, I'd go for a combobox instead.

larsoner:
> If only one option is selectable, you should use radio buttons.

I would, except 1) checkboxes are more standard in menus in my experience, and 2) *no option* is also an option. So checkboxes where at most 1 (but can be 0) are selectable is fine here I think.

I'd rather not use a combobox because then the text shows up, which I think is a waste of space (as it is for the other buttons). Which mode you're in is usually very obvious anyway from the actual display

cbrnr:
> I would, except 1) checkboxes are more standard in menus in my experience, and 2) no option is also an option. So checkboxes where at most 1 (but can be 0) are selectable is fine here I think.

I've never seen checkboxes where only 0 or 1 are selectable. Checkboxes are used for an arbitrary number of selections. I would use radio buttons and add a "None" (or "Hide"/"Off") option if you don't want a combobox.

larsoner:
> I've never seen checkboxes where only 0 or 1 are selectable.

I think I've seen this in menus before at some point (and I don't think I've ever seen combo boxes), but I'm not positive. Most things I went looking for online suggested a checkbox-with-exclusivity approach for menus, too. Even though the combobox approach involved a bit more complexity, I can ultimately live with it though (pushed now; the flickering is just some `peek` bug, it doesn't happen during actual interaction):

https://user-images.githubusercontent.com/2365790/161318271-4fec5922-9fa5-491a-a7ba-b87ed7326b4c.mp4

cbrnr:
I think this is now very nice! I like it!

cbrnr:
BTW, does the overview bar show annotations and/or events?

larsoner:
Not sure, feel free to check

agramfort:
works great !

agramfort (approved):

PR #91, opened 2022-04-04
ENH: Better HiDPI support

larsoner:
We should use logical pixels for line widths. My display has a physical-to-logical ratio of 1.777. The difference is subtle but it's there.

Before:

Screen Shot 2022-04-04 at 12 51 34 PM

After:

Screen Shot 2022-04-04 at 12 51 46 PM

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -285,7 +285,7 @@ class DataTrace(PlotCurveItem):
             else:
                 self.color = self.mne.ch_color_ref[self.ch_name]
 
-        self.setPen(_get_color(self.color, self.mne.dark))
+        self.setPen(self.mne.mkPen(_get_color(self.color, self.mne.dark)))
 
     @propagate_to_children
     def update_range_idx(self):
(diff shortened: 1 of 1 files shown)


larsoner:
@hoechenberger feel free to try this to see if it helps a little bit at least!

cbrnr:
Do we expose the line width so that people can set it to whatever they want?

larsoner:
Maybe at some point but not in this PR. That would be in MNE-Python first. I think either way now we can think about logical pixels instead of physical ones which is nice

hoechenberger:
I will take a look asap.

hoechenberger:
| `main`  | this PR |
| ------------- | ------------- |
| Screen Shot 2022-04-05 at 11 21 48  | Screen Shot 2022-04-05 at 11 21 03  |
| Screen Shot 2022-04-05 at 11 21 52  | Screen Shot 2022-04-05 at 11 21 19 |



@larsoner This is a good improvement already! I'd say we should merge this and if we decide we want to adjust the line width some more later on, we know where to look! Thanks!

hoechenberger (approved):

hoechenberger:
I cannot merge. Can somebody gibe me merge permissions please?

cbrnr:
But note that high frequencies disappear visually with increased line width. I'm fine with merging this, but we should expose the line width setting soon (maybe just provide three settings thin/medium/bold if we don't want to support arbitrary numbers).

hoechenberger:
> But note that high frequencies disappear visually with increased line width.

Yep, seems to be a tradeoff!

hoechenberger:
I haven't tried to really track this down but it appears that this PR also fixed the issue that the x axis label was partially cropped on my system. Nice!

PR #97, opened 2022-04-05
Fix x axis label text for Epochs (it used to say Time)

hoechenberger:
The text alignment is still wrong for me but at least the text itself is now correct for Epochs…

Screen Shot 2022-04-05 at 15 32 40

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3207,6 +3207,9 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 'qt_key': Qt.Key_Space
             }
         }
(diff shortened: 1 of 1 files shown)


larsoner (approved):

hoechenberger:
@larsoner I've changed it now such that the keyboard shortcut doesn't appear in the help either for Epochs
And when viewing raw data and toggling to "absolute" time, the x axis now has the label "Time of day"
WDYT?

hoechenberger:
> "absolute time", or "datetime" (maybe too much jargon),

I fact, I'd love to display the time zone as well, and then I'd be happy with "Absolute time". But let's keep this for later, yes

hoechenberger:
Good to go from my end

larsoner:
Thanks @hoechenberger !

PR #98, opened 2022-04-05
Make MEG channel order in butterfly plot match order in overview bar

### LINKED ISSUE #94: Channel type order in butterfly plot doesn't match order in overview bar / order in non-butterfly mode
MWE:

```python
# %%
import mne

sample_dir = mne.datasets.sample.data_path()
sample_fname = sample_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(sample_fname)
raw.crop(tmax=60)
raw.plot()
```

We can see channels are ordered as: grad, mag, eeg:
Screen Shot 2022-04-05 at 11 40 37

But in butterfly mode, the order is mag, grad, eeg:
Screen Shot 2022-04-05 at 11 40 52

This should be harmonized.

hoechenberger:
Fixes #94

MWE:
```python
# %%
import mne


sample_dir = mne.datasets.sample.data_path()
sample_fname = sample_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(sample_fname, preload=True)
raw.crop(tmax=60)
raw.plot(block=True, butterfly=True)
```

| before | after |
|----|---|
| Screen Shot 2022-04-05 at 20 10 10 | Screen Shot 2022-04-05 at 20 09 56 |

cc @agramfort

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -69,6 +69,12 @@ except ImportError:
 
 name = 'pyqtgraph'
 
+# MNE's butterfly plots traditionally default to the channel ordering of
+# mag, grad, ..., which is inconsistent with the order in non-butterfly mode
+# and hence doesn't match the order in the overview bar either. So we swap
+# grads and mags here.
+DATA_CH_TYPES_ORDER = ('grad', 'mag', *_DATA_CH_TYPES_ORDER_DEFAULT[2:])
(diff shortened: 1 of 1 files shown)


agramfort (approved):

agramfort:
thx @hoechenberger

PR #99, opened 2022-04-05
Remove bits of code that are not needed with MNE-Python 1.0 anymore

hoechenberger:


hoechenberger:
Oh, do we still support MNE 0.24..?

agramfort:
I would say it's fine now to require mne 1.0 but you never know. It does not cost much to keep this code and maybe keep a few users happy...

hoechenberger:
Ok so I'm going to close this for now

PR #101, opened 2022-04-05
Always keep projectors window on top

hoechenberger:
Also minor change for code health to enforce more explicit parameter specification when creation projection dialog window.

MWE:
```python
# %%
import mne


sample_dir = mne.datasets.sample.data_path()
sample_fname = sample_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(sample_fname, preload=True)
raw.crop(tmax=60)
raw.plot(block=True)
```

https://user-images.githubusercontent.com/2046265/161846060-dc4808b3-1f7c-45ee-969e-66d85657e6c9.mov

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1621,7 +1621,7 @@ class HelpDialog(_BaseDialog):
     """Shows all keyboard-shortcuts."""
 
     def __init__(self, main, **kwargs):
-        super().__init__(main, **kwargs)
+        super().__init__(main, title='Help', **kwargs)
 
         # Show all keyboard-shortcuts in a Scroll-Area
         layout = QVBoxLayout()
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1695,10 +1695,10 @@ class HelpDialog(_BaseDialog):
 class ProjDialog(_BaseDialog):
     """A dialog to toggle projections."""
 
-    def __init__(self, main, **kwargs):
+    def __init__(self, main, *, name):
         self.external_change = True
         # Create projection-layout
-        super().__init__(main, **kwargs)
+        super().__init__(main, name=name)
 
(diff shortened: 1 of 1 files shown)


hoechenberger:
I now added a window title too:

larsoner (changes requested):
On linux at least, this causes the dialog to be on top *system-wide*:

Screenshot from 2022-04-05 16-55-24



Maybe there is some hint here about parenting that would fix it?

https://stackoverflow.com/questions/29568038/qt-window-that-stays-on-top-of-the-application-but-not-other-windows

FYI, for me on Linux the behavior is already correct on `main`.

hoechenberger:
> On linux at least, this causes the dialog to be on top _system-wide_:

Same on macOS, but it was the only quick solution I could find so far for a non-modal window. Already a great improvement for me, though…



> https://stackoverflow.com/questions/29568038/qt-window-that-stays-on-top-of-the-application-but-not-other-windows

I read that, and we already have a parent, but this doesn't seem to suffice on macOS

> FYI, for me on Linux the behavior is already correct on `main`.

Not for me on macOS, unfortunately :(

larsoner:
Pushed a fix that works on macOS (and actually makes the title bar smaller, which is nice!), plus generalized to all `_BaseDialog`s, plus a fix for
```
TypeError: SettingsDialog.closeEvent() takes 1 positional argument but 2 were given
```
Let's see if it works on Linux, too...

hoechenberger:
Great, works for me!

Re problem with the settings dialog, I think quite some effort has been put into this in #85, so I'd suggest to revert this change here and merge #85 instead to give the author credit, no?

larsoner:
> I'd suggest to revert this change here

Done, I'll merge this PR once green now

hoechenberger:
Thank you, @larsoner!

PS: Can you please give me permissions to merge PRs myself here too? Otherwise I'll always have to wait for you 😅

PR #102, opened 2022-04-06
CI: Bump to maint/1.0

larsoner:
Bump to use maint/1.0 of MNE, but keep one 0.24 build around. If it's not too painful to maintain backward compat to two versions, I think we should.

Also attempts to bump the main build matrix to 3.10, let's see if it works.


larsoner:
To get to 3.10, I removed the VTK requirement. I think we can rely on gl-ci-helpers (and `glxinfo`, which I've now added) to make sure we have a working OpenGL version.

hoechenberger (approved):

larsoner:
Ready for review/merge from my end

PR #104, opened 2022-04-07
MNT: Add support for QtPy

GuillaumeFavelier:
This PR adds support for `qtpy`.

* `QDesktopWidget` and `QApplication.desktop()` are deprecated in `Qt 5` and removed in `Qt 6` so I migrated to `QScreen`: https://doc.qt.io/qt-6/widgets-changes-qt6.html#qdesktopwidget-and-qapplication-desktop
* Similarly to `vtk`, `pyqtgraph` checks Qt APIs by trying to import the package directly instead of looking for actual features : https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt/__init__.py#L31-L51
  To fix this locally, I use the `PYQTGRAPH_QT_LIB` variable.

TODO:

- [x] Make everything (important) work on all Qt platforms
- [x] Merge https://github.com/mne-tools/mne-python/pull/10509

Then this PR is ready to go


larsoner:
We should also port over the Qt job from MNE-Python https://github.com/mne-tools/mne-python/blob/main/azure-pipelines.yml#L147

mne_qt_browser/_pg_figure.py
```diff
@@ -20,22 +20,22 @@
...
-                             QGridLayout, QHBoxLayout, QInputDialog,
-                             QLabel, QMainWindow, QMessageBox, QToolButton,
-                             QPushButton, QScrollBar, QWidget, QMenu,
-                             QStyleOptionSlider, QStyle, QActionGroup,
-                             QApplication, QGraphicsView, QProgressBar,
-                             QVBoxLayout, QLineEdit, QCheckBox, QScrollArea,
-                             QGraphicsLineItem, QGraphicsScene, QTextEdit,
-                             QSizePolicy, QSpinBox, QDesktopWidget, QSlider,
-                             QWidgetAction, QRadioButton)
+from qtpy.QtCore import (QEvent, QThread, Qt, Signal, QRectF, QLineF,
```
larsoner, line 23: And I guess we'll want to add `qtpy` to `setup.py:install_requires`

larsoner:
... actually all the CI jobs here are GH actions, so I would just add a new entry in `qt_viz_tests.yml`

larsoner:
Also keep in mind you'll want to rebase or merge with `main` because another PyQt5 import was added in the now merged https://github.com/mne-tools/mne-qt-browser/pull/85/files#diff-835cae3833968c10d566cdfb4f83de733e1a89736033836229493057266e1c2fR7

larsoner:
I pushed a commit that at least made things work for me on PyQt5 again -- `widget.screen()` is only available on 5.14+.

On PySide6 I get this ugly error:
```
(qt6) $ mne browse_raw ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif 
...
File "/Users/larsoner/python/mne-python/mne/viz/_figure.py", line 613, in _load_backend
    from mne_qt_browser import _pg_figure as backend
  File "/Users/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 2620, in <module>
    class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
  File "/Users/larsoner/mambaforge/envs/qt6/lib/python3.9/abc.py", line 106, in __new__
    cls = super().__new__(mcls, name, bases, namespace, **kwargs)
TypeError: Shiboken.ObjectType.__new__(_PGMetaClass) is not safe, use type.__new__()
```
It has something to do with the _PGMetaClass stuff, not sure how to fix it :(

.github/workflows/qt_viz_tests.yml
```diff
@@ -167,10 +151,6 @@ jobs:
         run: |
           pip uninstall -yq PyQt5 PyQt5-Qt5 PyQt5-sip
           pip install ${{ matrix.qt }}
-          echo "::set-env name=PYQTGRAPH_QT_LIB::${{ matrix.qt }}"
```
larsoner: FYI @GuillaumeFavelier I removed this as it should not be necessary -- PyQtGraph says it first looks to see which APIs have been imported, and since our `qtpy` import is above the `pyqtgraph` import, the resulting qt api should be respected

larsoner:
Wow, green! Locally when trying any keyboard shortcut, though, I get things like:
```
  File "/Users/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 4331, in keyPressEvent
    'Ctrl': '4' in hex(int(event.modifiers())),
TypeError: int() argument must be a string, a bytes-like object or a number, not 'KeyboardModifier'
  File "/Users/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 4331, in keyPressEvent
    'Ctrl': '4' in hex(int(event.modifiers())),
TypeError: int() argument must be a string, a bytes-like object or a number, not 'KeyboardModifier'
```
@GuillaumeFavelier I'll stop here for now, feel free to continue when you want :)

larsoner:
Modifier bugs were present in `main` actually, so I fixed them in https://github.com/mne-tools/mne-qt-browser/pull/109

larsoner:
Pushed another little commit to take care of the warnings due to using `'AnyStyle'`:
```
qt.qpa.fonts: Populating font family aliases took 198 ms. Replace uses of missing font family "AnyStyle" with one that exists to avoid this cost. 
```

larsoner:
FYI I need https://github.com/mne-tools/mne-python/pull/10508 for macOS + Python3.10. I'll see if I can replicate on CIs...

larsoner:
Okay, PySide6, PyQt6, and PyQt5 all work on my macOS system!

larsoner:
Setting back to WIP because, even though it's getting close, https://github.com/mne-tools/mne-python/pull/10509 should go in first (which should also go in after https://github.com/mne-tools/mne-python/pull/10500)

mne_qt_browser/_pg_figure.py
```diff
@@ -390,13 +390,12 @@ def update_data(self):
...
             data = self.mne.data[self.range_idx]
+        times = self.mne.times
 
         # Get decim-specific time if enabled
         if self.mne.decim != 1:
-            times = self.mne.decim_times[self.mne.decim_data[self.range_idx]]
+            times = times[::self.mne.decim_data[self.range_idx]]
             data = data[..., ::self.mne.decim_data[self.range_idx]]
-        else:
-            times = self.mne.times
```
larsoner, line 399: This "caching" of the decimated times was causing problems with tests locally. Creating views (which slicing does) is a `ns`-order operation:
```
In [1]: import numpy as np
In [2]: x = np.random.RandomState(0).randn(10000)
In [3]: %timeit x[100:1111]
136 ns ± 3.88 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
So let's not bother, and instead just slice times at the same time / way as the data

larsoner:
Okay @GuillaumeFavelier I think I'm actually done now :)

Feel free to review sometime next week and we can get this in once https://github.com/mne-tools/mne-python/pull/10509 is merged

larsoner:
@agramfort or @hoechenberger feel free to review and merge if it works for you on PyQt6 and/or PySide6

agramfort (approved):

agramfort:
thx @GuillaumeFavelier @larsoner it works like a charm !

PR #105, opened 2022-04-07
ENH: Allow passing overview_mode

larsoner:
1. Check `self.mne.overview_mode` for validity
2. Support MNE_BROWSER_OVERVIEW_MODE in MNE-Python (https://github.com/mne-tools/mne-python/pull/10501)
3. Ensure the splash screen is always closed, including when there is an error in `__init__`
4. Remove `self.mne.show_overview_bar` in favor of just using `self.mne.overview_bar.isVisible()` (let Qt keep track instead of us)

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -113,13 +113,6 @@ jobs:
         if: runner.os == 'Linux' && contains(matrix.opengl, 'opengl')
       - name: Show system information
         run: mne sys_info
-      # This workaround can be removed once https://github.com/mne-tools/mne-python/pull/10502
-      # is merged and backported, and MNE-Python >= 1.0.1 is released
-      - name: Protect the user config
-        run: |
-          set -e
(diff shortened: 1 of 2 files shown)


larsoner:
This should pass once https://github.com/mne-tools/mne-python/pull/10502 is merged. I'll probably disable the macOS maint/1.0 run for now (or mark it as an allowed failure somehow) until we backport and cut a maintenance release 1.0.1 of MNE-Python.

larsoner:
Okay I found a nice little workaround with a note about when we can remove it, no need to wait for https://github.com/mne-tools/mne-python/pull/10502

All green, ready for review/merge from my end

larsoner:
@cbrnr @hoechenberger  feel free to review/merge this and https://github.com/mne-tools/mne-python/pull/10501 if you're happy.

FYI I changed the behavior of the overview bar a bit with a visibility toggle rather than a "hidden" entry. Really there are two states to track: the exclusive (radiobutton) mode selection between channels/zscore/empty, and the binary (checkbox) visibility of the overview bar. This works better with the visibility keyboard toggle, where it should effectively stay in channels/zscore/empty mode when being hidden/unhidden.

PR #107, opened 2022-04-07
BUG: Fix the pixel ratio

larsoner:
Directly use the `devicePixelRatio` instead of the (incorrect) self-computation. On my system the old value was ~1.7 now it's the correct 2.

@hoechenberger feel free to review/merge if you're happy


larsoner:
This one is a trivial fix for what was wrong before so I'll put it in

Issue #83, opened 2022-03-30
Improve dark mode


mscheltienne:
Earlier today, on an env with the latest dev version of `mne` and `mne_qt_browser` on my MacBook, the `raw.plot()` I opened were in dark mode/theme. And sure I do like dark mode usually, but here it was really not that great and made the examination of the signal a pain. 

**Feature request**: 
- [ ] in the setting window, a toggle for light/dark theme.
OR
- [ ] in the setting window, a drop-down list to select the theme.

**Bug**:
- [ ] on macOS, with dark mode enabled at the system level, the argument `thene='light'` in `raw.plot()` is ignored.

----------

I think the MacBook is set up in dark mode at the system-level, which is not the case on the Windows PC I'm using at the moment where the `raw.plot()` do not display with a dark theme.

hoechenberger:
> but here it was really not that great and made the examination of the signal a pain.

Could you explicate what was the issue specifically?

mscheltienne:
I'll make a couple of screenshots tomorrow, traces had a ~white-ish color while the background was a very dark black.. 
It's not really an issue, as I am sure some people love it this way; but for me, it wasn't a theme I enjoyed using.

larsoner:
In the meantime there is a `theme` arg to raw.plot, along with a `MNE_BROWSER_THEME` config var

https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.plot
https://github.com/mne-tools/mne-python/pull/10418

But I agree we should keep this open to allow toggling it in the browser itself, maybe via a toolbar button or something

hoechenberger:
> I'll make a couple of screenshots tomorrow, traces had a ~white-ish color while the background was a very dark black.. 
> 
> It's not really an issue, as I am sure some people love it this way; but for me, it wasn't a theme I enjoyed using.

I was mostly asking because it could be that we need to adjust the theme to make accessibility better. Could be that we need to adjust colors or line width or so to reach a reasonable contrast.

mscheltienne:
Oh, I completely missed this `theme` argument! Thanks!
However, it does not work 😅 and the theme remains dark as on the screenshot below.

So starting with a 64 EEG recordings, with large annotations:

![Screenshot 2022-03-31 at 11 11 48](https://user-images.githubusercontent.com/73893616/161020454-4d28fed1-6749-473b-ae88-bba455a48a32.png)

Then with a sample dataset from MNe including MEG / EEG with no annotations:

[...]

PR #108, opened 2022-04-08
ENH: Better dark colors

larsoner:
This at least sets up a framework by which we can set the dark mode analogs for our light-mode colors:

| main-dark | main/PR-light | PR-dark |
| -- | -- | -- |
| Screen Shot 2022-04-08 at 10 36 11 AM | Screen Shot 2022-04-08 at 10 39 02 AM | Screen Shot 2022-04-08 at 10 36 26 AM |

For now I tried to choose similar colors based on the matplotlib list.

@mscheltienne you mentioned the colors not being so good, is this any better?

Helps with #83

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -138,7 +138,7 @@ except ImportError:
 # https://matplotlib.org/3.1.0/gallery/color/named_colors.html
 _dark_dict = {
     # 'w' (bgcolor)
-    (255, 255, 255): (15, 15, 15),  # chrome's dark mode bgcolor
+    (255, 255, 255): (35, 35, 35),  # chrome's dark mode bgcolor
     # 'k' (eeg, eog, emg, misc, stim, resp, chpi, exci, ias, syst, dipole, gof,
     #      bio, ecog, fnirs_*, csd, whitened)
     (0, 0, 0): (255, 255, 255),   # 'w'
@@ -165,6 +165,7 @@ _dark_dict = {
 
 def _get_color(color_spec, invert=False):
     """Wraps mkColor to accept all possible matplotlib color-specifiers."""
+    orig_spec = color_spec
     try:
         # Convert matplotlib color-names if possible
         color_spec = _to_rgb(color_spec, alpha=True)
@@ -188,7 +189,7 @@ def _get_color(color_spec, invert=False):
         if key[:3] in _dark_dict:
             color.setRgb(*(_dark_dict[key[:3]] + key[-1:]))
         else:
-            print(f'Missed {key} ({type(key)})')
+            logger.debug(f'Missed {key} from {orig_spec}')
             rgba = np.array(color.getRgbF())
             lab = cspace_convert(rgba[:3], 'sRGB1', 'CIELab')
             lab[0] = 100. - lab[0]
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -134,6 +134,35 @@ except ImportError:
         return app
 
 
+# Mostly chosen manually from
+# https://matplotlib.org/3.1.0/gallery/color/named_colors.html
+_dark_dict = {
+    # 'w' (bgcolor)
+    (255, 255, 255): (15, 15, 15),  # chrome's dark mode bgcolor
+    # 'k' (eeg, eog, emg, misc, stim, resp, chpi, exci, ias, syst, dipole, gof,
+    #      bio, ecog, fnirs_*, csd, whitened)
+    (0, 0, 0): (255, 255, 255),   # 'w'
+    # 'darkblue' (mag)
+    (0, 0, 139): (173, 216, 230),  # 'lightblue'
+    # 'b' (grad, hbr)
+    (0, 0, 255): (100, 149, 237),  # 'cornflowerblue'
+    # 'steelblue' (ref_meg)
+    (70, 130, 180): (176, 196, 222),  # 'lightsteelblue'
+    # 'm' (ecg)
+    (191, 0, 191): (238, 130, 238),  # 'violet'
+    # 'saddlebrown' (seeg)
+    (139, 69, 19): (244, 164, 96),  # 'sandybrown'
+    # 'seagreen' (dbs)
+    (46, 139,  87): (32, 178, 170),  # 'lightseagreen'
+    # '#AA3377' (hbo), closest to 'mediumvioletred'
+    (170, 51, 119): (255, 105, 180),  # 'hotpink'
+    # 'lightgray' (bad_color)
+    (211, 211, 211): (105, 105, 105),  # 'dimgray'
+    # 'cyan' (event_color)
+    (0, 255, 255): (0, 139, 139),  # 'darkcyan'
+}
+
+
 def _get_color(color_spec, invert=False):
     """Wraps mkColor to accept all possible matplotlib color-specifiers."""
     try:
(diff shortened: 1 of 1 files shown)


larsoner:
@hoechenberger @agramfort @cbrnr you also had comments in https://github.com/mne-tools/mne-qt-browser/issues/83#issuecomment-1086865664 for example, this moves us toward a custom set of colors

larsoner:
And here is the script I used to test:
```Python
import numpy as np
import mne
ch_t = ('grad', 'mag', 'ref_meg', 'eeg', 'eeg', 'ecg', 'seeg', 'dbs', 'hbo')
ch_n = list(ch_t)
ch_n[4] = 'bad'
info = mne.create_info(ch_n, 1000., ch_t)
info['bads'] = ['bad']
data = np.random.RandomState(0).randn(len(ch_t), 1000) * 0.1
raw = mne.io.RawArray(data, info)
events = np.array([[200, 0, 1], [300, 0, 2]])
annot = mne.Annotations([0.5, 0.8], 0.1, ['BAD_Segment', 'Test'])
raw.set_annotations(annot)
raw.plot(scalings=1., block=True, order=np.arange(len(ch_t)), events=events)
```
It can be launched with `MNE_BROWSER_THEME=dark python ...` for example to run in dark mode.

mscheltienne:
It looks better for me, at least in terms of esthetic and viewing experience, but I also mentioned being color-blind.. so I'm not the best to judge colors 😅 I think the background of the plotting area is still a bit too dark for me. I am used to a dark mode with a background color closer to the toolbar background (similar to macOS, Firefox dark modes).
However, it did become a bit harder to differentiate some of the channel types by color (most likely because they are in the range of my color-blindness), but I don't think this is a big issue as I usually know what I'm plotting and can always refer to the channel names.

larsoner:
> I think the background of the plotting area is still a bit too dark for me. I am used to a dark mode with a background color closer to the toolbar background (similar to macOS, Firefox dark modes).

I tried increasing it a bit to the `(53, 53, 53)` that Chrome uses when you open an empty tab, but this seemed too light. The value I have now (`(17, 17, 17)`) is from the dark part of Gmail's dark mode (or maybe it's some theme I chose?).

Want to try different values @mscheltienne and propose what you think is best? Then maybe I can try and we can agree on something in between (or your value if it looks okay already!)

mscheltienne:
I'm exhausted tonight, but I'll have a look this weekend if I can tune a bit more your values.
I don't think a consensus can be reached that satisfies everyone anyway, but let's try and propose a couple of options :)

I'll try with the firefox colors: https://design.firefox.com/photon/visuals/color.html

larsoner:
1. Here is (50, 50, 50), which is the widget bgcolor:

    Screen Shot 2022-04-08 at 4 33 13 PM

2. If we split the difference and use `(35, 35, 35)` I think it looks decent:

    Screen Shot 2022-04-08 at 4 33 39 PM

3. A bit darker `(25, 25, 25)`:
    Screen Shot 2022-04-08 at 4 37 02 PM

4. And just for completeness, this is what I had before (which was actually `(15, 15, 15)`), which I don't like as much as 25 (or maybe 35):

    Screen Shot 2022-04-08 at 4 35 53 PM

I think (2) and (3) (35 and 25) are my favorites actually

agramfort:
2 or 3 work for me 

Thx 🙏

cbrnr:
I like (3).

mscheltienne:
From the 4 you propose, I prefer (2) or (3).

I left my macOS at work, so I'm now using Windows, and I only now realize there are large differences between macOS and Windows dark-mode. Any reason why this is the case?

I gave a shot to 5 colors for the background:
- (50, 50, 50)
- (12, 12, 13) -- Mozilla Grey 90
- (42, 42, 46) -- Mozilla Grey 80
- (56, 56, 61) -- Mozilla Grey 70
- (74, 74, 79) -- Mozilla Grey 60

Capture

Note that they chose to slightly increase the blue proportion compared to the 2 other colors. 


| `(50, 50, 50)` | `(42, 42, 46)` | `(12, 12, 13)` | `(56, 56, 61)` | `(74, 74, 79)` | 
| :-------: | :------: | :------: | :------: | :------: |
| ![v1][1] | ![v2][2] | ![v3][3] | ![v4][4] | ![v5][5] |


[1]: https://user-images.githubusercontent.com/73893616/162568032-a37b5f48-359d-441f-9ab4-b02d5ba5afb4.png
[2]: https://user-images.githubusercontent.com/73893616/162568182-935b09e2-9d88-4068-a4b2-0a3a834b0a71.png
[3]: https://user-images.githubusercontent.com/73893616/162568069-ff3029ae-59df-4f9d-9ca1-c582bbb938c4.png
[4]: https://user-images.githubusercontent.com/73893616/162568075-4b6ce60a-7bce-4dac-861c-2c2f4e597a7c.png
[5]: https://user-images.githubusercontent.com/73893616/162568087-3a4f1054-0202-4d88-9a53-726722a1882d.png

Here is the Grey 80 with the entire window:

image

Personally, I like how the viewing/plotting area looks in this one, especially the contrast between the background on the edge and the lighter plotting area background + the contour applied to the plotting area.
However, the toolbar and status bar looks bad.

larsoner:
> I left my macOS at work, so I'm now using Windows, and I only now realize there are large differences between macOS and Windows dark-mode. Any reason why this is the case?

Yes, on macOS (at least with https://github.com/mne-tools/mne-python/pull/10500) it uses the native styling provided by Qt, which is similar to what macOS uses. AFAIK there isn't native dark styling for Qt in Windows, so we have to use a qdarkstyle stylesheet for it. You will also experience this theme on macOS if you have light mode for your system, but ask for a dark theme. This stylesheet has a blueish tint.

> Note that they chose to slightly increase the blue proportion compared to the 2 other colors.

Interesting. I looked at the bgcolor of the builtin "Digital Color Meter.app" I had been using to pull colors, and it actually uses `(35, 31, 41)` (I had my mouse over its background), but this ends up looking noticeably purple to me:

| Meter | MNE |
| -- | -- |
| Screen Shot 2022-04-09 at 8 44 21 AM | Screen Shot 2022-04-09 at 8 59 00 AM |

And the left pane of my Finder is `(41, 41, 51)`. So some blue shift seems pretty common in macOS dark theming.

> Personally, I like how the viewing/plotting area looks in this one, especially the contrast between the background on the edge and the lighter plotting area background + the contour applied to the plotting area.

I agree `(42, 42, 46)` looks decent. On macOS though it comes a bit too close to the other colors I think:

Screen Shot 2022-04-09 at 8 52 33 AM

It looks like Safari actually uses `(30, 30, 30)` for its center here:

Screen Shot 2022-04-09 at 8 55 05 AM

And this is what it would look like for us:

| macOS native | macOS darkdetect-themed |
| -- | -- |
| Screen Shot 2022-04-09 at 9 03 28 AM | Screen Shot 2022-04-09 at 9 04 56 AM |

Given that it splits the difference between the (2) and (3) I had and people liked above, is used by Apple in one of their premier applications, won't create a hue issue on macOS, and I think is different enough from the darkdetect blue to not be seen as an error, I'm inclined to go with that one.

mscheltienne:
Ok, that makes sense now. It's a bit weird that the browser has 2 dark-theme on macOS.
So yes, anything too close to `(50, 50, 50)` will not work on macOS. `(30, 30, 30)` looks good.

larsoner:
>. It's a bit weird that the browser has 2 dark-theme on macOS invoked by the same theme='dark' depending on installed dependencies.

What I'm talking about assumes `qdarkstyle` is installed in both cases. It's really whether the *system* is in dark mode or light mode that changes things. If it's in dark mode and you request dark mode, we can use Qt's native styling (gray). If the system is in light mode but you request dark mode, we have to overwrite the stylesheets to get dark mode, which means using qdarkstyle (blueish). 

The same thing actually happens when your macOS system is in *dark* mode but you request *light* mode -- we have to use `qdarkstyle` to light-theme thingns. The differences there are more subtle though:

| light: system in light (Qt native) | light: system in dark (qdarkstyle) |
| -- | -- |
| Screen Shot 2022-04-09 at 9 47 56 AM | Screen Shot 2022-04-09 at 9 47 39 AM |

Sounds like we've converged on 30, so I'll go ahead and merge. We can always tweak trace colors further if people try the ones I've chosen here and find them to be unusable!

PR #109, opened 2022-04-08
BUG: Modifiers

larsoner:
Fixes shift-arrow behavior for me on macOS. Also it will fix the bug with https://github.com/mne-tools/mne-qt-browser/pull/104

I'm going to go double-check that this works on Windows and Linux.


larsoner:
Checked that now shift works on all three OSes. It actually works on `main` on Windows and Linux for me, but this PR does fix macOS and simplifies the code I think.

Ready for review/merge from my end

agramfort (approved):

PR #111, opened 2022-04-09
Add theme related req to requirements.txt

mscheltienne:
`darkdetect` and `qdarkstyle` are 2 optional dependencies for theme management.
I think they are used by `mne-python` before invoking the browser, but since theme selection is a key feature of the browser, how about adding them as a requirement here as well? This way, a base install of `mne` will include them by default. 

If they are used by other GUI through `mne-python`, then maybe it would be worth adding them to one of the `requirements_extra.txt` of `mne-python` as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,11 +2,12 @@
 numpy
 scipy
 matplotlib
-mne
+PyQt5>=5.12
 qtpy
 scooby
-PyQt5
-pyqtgraph
+qtpy
+pyqtgraph>=0.12.3
+colorspacious
 pyopengl; platform_system=="Darwin"
 darkdetect
 qdarkstyle
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ PyQt5>=5.12
 qtpy
 scooby
 qtpy
+mne>=0.24
 pyqtgraph>=0.12.3
 colorspacious
 pyopengl; platform_system=="Darwin"
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,5 @@ scooby
 PyQt5
 pyqtgraph
 pyopengl; platform_system=="Darwin"
+darkdetect
+qdarkstyle
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ matplotlib
 PyQt5>=5.12
 qtpy
 scooby
-qtpy
 mne>=0.24
 pyqtgraph>=0.12.3
 colorspacious
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ def parse_requirements_file(fname):
             requirements.append(req)
     return requirements
 
+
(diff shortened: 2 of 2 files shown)


hoechenberger:
I also think it should be hard requirements and therefore go to setup.py or .cfg as well

mscheltienne:
Yep, I only now realize that the `setup.py` does not read the requirements from the `.txt` file. Is this on purpose?

larsoner:
> Is this on purpose?

It's possible that requirements.txt includes/included all optional dependencies, like the `requirements.txt` in MNE-Python. After adding qdarkstyle and darkdetect, though, there might not be any more optional dependencies (other than PyOpenGL, which is handled via an `[opengl]` option already)

mscheltienne:
Actually, it was missing some dependencies and it was missing version limitations. So I guess there is no point in not loading the dependencies from the `.txt` file. I added the parser from MNE.

requirements.txt
```diff
@@ -2,9 +2,12 @@
 numpy
 scipy
 matplotlib
-mne
```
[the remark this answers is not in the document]
mscheltienne: typo.. I wanted to add the minimum version, not remove it. Not sure what I did to get this result.
I double-checked, the others should be fine now.

requirements.txt
```diff
@@ -2,9 +2,13 @@
...
+PyQt5>=5.12
 qtpy
 scooby
-PyQt5
-pyqtgraph
+qtpy
```
larsoner, line 8: qtpy appears twice

larsoner, line 8:
```suggestion
```

mscheltienne: so much for the double-check.. I did look at this file for a solid minute 🙈 and still missed it. Thanks!

larsoner:
Thanks @mscheltienne !

hoechenberger:
We don't we drop requirements.txt instead?

IMHO we're not making proper use of it anyway: it's supposed to specify exact versions to be used for a specific deployment, not just a list of deps with loose version constraints (or none at all). This is what should go to setup.py 

As it is now, one has to manually "parse" the code that imports requirement.txt in setup.py to figure which are the actual dependencies of this package...

larsoner:
> IMHO we're not making proper use of it anyway: it's supposed to specify exact versions to be used for a specific deployment, not just a list of deps with loose version constraints (or none at all). 

You mean static pins? Or the whole dependency tree (including dependencies of dependencies of ...)? This actually hasn't been how I've seen it used most places

https://github.com/pandas-dev/pandas/blob/main/requirements-dev.txt
https://github.com/scipy/scipy/blob/main/doc_requirements.txt

And it's not how we've used it in MNE. I'd rather stay consistent with how we've used it in MNE as much as possible.

larsoner:
> We don't we drop requirements.txt instead?

As a dev I've found it useful multiple times to have a `requirements.txt` that I could `pip -r` with first. It's also a nice standard format that's easy to read (and parse, etc.).

PR #116, opened 2022-04-19
ENH: Respect activation

larsoner:
1. Allow `_qt_raise_window` to actually control window-raising behavior
2. If you click away from mne-qt-browser, then back to it using a child dialog, the main window should be raised. This can be tested locally with:
    ```
    import mne
    raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() / 'MEG' / 'sample' / 'sample_audvis_raw.fif')
    raw.plot(group_by='selection', block=True)
    ```
3. Move backported functionality to `mne_qt_browser/_fixes.py`
4. No need to `self.mne.splash.close()` in `show` as this will be handled already by `_safe_splash`

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+branch = True
+source = mne_qt_browser
+omit =
+    */setup.py
+    */mne_qt_browser/_fixes.py
+    */mne_qt_browser/conftest.py
--- a/mne_qt_browser/_fixes.py
+++ b/mne_qt_browser/_fixes.py
@@ -1,3 +1,16 @@
+# -*- coding: utf-8 -*-
+"""Backports for MNE versions (mostly)."""
+
+# Author: Martin Schulz <dev@earthman-music.de>
+#
+# License: BSD-3-Clause
+
+from contextlib import contextmanager
+import platform
+import sys
+
+from mne.utils import logger
+
 ###############################################################################
 # MNE 1.0+
 try:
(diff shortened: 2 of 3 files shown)


larsoner:
@hoechenberger feel free to review and merge if you're happy

hoechenberger:
Code looks good to me, but I cannot test until tomorrow, so if this is urgent, go ahead and merge; otherwise I can do it tomorrow :)

hoechenberger:
@larsoner Actually I realized that neither in this branch nor in `main` the child dialog remains open when I switch away to a different application window on macOS – truly unusual behavior! Meaning I cannot test the changes you're implementing here

https://user-images.githubusercontent.com/2046265/164183419-03443108-791b-4a4e-aabc-dd445f6e9de7.mov

larsoner:
> truly unusual behavior! Meaning I cannot test the changes you're implementing here

This seems like an intentional design choice by the Qt folks given [you have to opt in to keeping the tool window visible on macOS](https://doc.qt.io/qt-5/qt.html#WindowType-enum):

Qt::WA_MacAlwaysShowToolWindow | 96 | On macOS, show the tool window even when the application is not active. By default, all tool windows are hidden when the application is inactive.
-- | -- | --

We could add this separately I guess, but I'm not sure we want to deviate from the default here?

larsoner:
I pushed, feel free to try it now to see if it seems more natural

hoechenberger:
@larsoner I was mistaken, I just checked with a stock macOS app (Preview.app), and indeed the tool windows disappear when the application loses focus. So the behavior I complained about is, in fact, the way it should be!

hoechenberger:
Thank you, @larsoner!

PR #118, opened 2022-04-30
[BUG]: Close annotations-mode when pressing escape

### LINKED ISSUE #74: When pressing ESC in annotations mode, exit annotations mode, do not close the browser
Currently, pressing ESC will exit the browser window. However, when I'm in annotations mode, my intuition is that I'd exit this mode, and only after an additional press of ESC the browser would quit.

The current behavior made me quit the browser several times now by accident…

marsipu:
#### Reference issue
Fixes #74.

This closes annotations-mode when pressing escape.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3248,7 +3248,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             },
             'escape': {
                 'qt_key': Qt.Key_Escape,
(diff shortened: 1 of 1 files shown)


hoechenberger:
Works great for me! The only issue I had was that I accidentally pressed ESC twice one time, and then the main window closed too :( 

Do you think we could add a "do you really want to close?" dialog to the main window?
Possibly with a "don't ask me again" checkbox :)

But this is a different issue from this PR

hoechenberger:
Thank you, @marsipu!

PR #120, opened 2022-05-12
fix lasso

marsipu:
This fixes the following error message when drawing a lasso in the selection-dialog.
`AttributeError: '_ChannelFig' object has no attribute 'mne'`

#### Steps to reproduce
`raw.plot(group_by='selection')`


larsoner:
Thanks @marsipu !

larsoner:
Do you want to cut a 0.3.1 with these fixes?

marsipu:
> Do you want to cut a 0.3.1 with these fixes?

Yes, good idea

PR #124, opened 2022-06-18
BUG: Fix crosshair

marsipu:
This fixes a small bug when using the crosshair (activated by pressing `x`).


agramfort (approved):
maybe add a what's new entry to document the bug fix?

larsoner:
I think so far there is no `latest.inc` for this repo and changes just end up in the Releases page, auto-generated. Though I suppose you could add an entry to MNE-Python!

hoechenberger:
> changes just end up in the Releases page, auto-generated

I would therefore encourage everyone to be a bit more verbose when it comes to PR titles, as this is what will show up in the auto-generated changelog. Neither from the title of this PR nor from any comments here can I deduce what was even the problem 😅

PR #126, opened 2022-06-18
WIP: Add a menu

hoechenberger:
This is WIP to add a menu for better discoverability of functionality, and to make things more accessible and feel more like a real application.

Qt does some magic to move the menu entries that match certain patterns into the right place on macOS:
https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QMenuBar.html#qmenubar-as-a-global-menu-bar

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,4 @@ dmypy.json
 
 *.orig
 junit-results.xml
+.vscode/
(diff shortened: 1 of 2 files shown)


hoechenberger:
I've added a Help menu and this comes with that macOS-typical discoverability feature for free:


https://user-images.githubusercontent.com/2046265/174440356-29cc3581-6727-4eb7-a083-0f1dec569c6b.mov

PR #139, opened 2022-08-14
remove add/remove objects for view-range

### LINKED ISSUE #82: Bug with annotation display: AnnotRegion are deleted?
Running the latest dev version of `mne` and `mne_qt_browser`, I got a bug with annotations. The file used in the video below has some annotations setup and when scrolling I get the following error in the console:

```
>>> Traceback (most recent call last):
  File "c:\users\mathieu\documents\git\mne-qt-browser\mne_qt_browser\_pg_figure.py", line 3518, in _xrange_changed
    self._update_annotations_xrange(xrange)
  File "c:\users\mathieu\documents\git\mne-qt-browser\mne_qt_browser\_pg_figure.py", line 3570, in _update_annotations_xrange
    self.mne.plt.addItem(region)
  File "C:\Users\Mathieu\Documents\pyvenv\mne\lib\site-packages\pyqtgraph\graphicsItems\PlotItem\PlotItem.py", line 535, in addItem
    self.vb.addItem(item, *args, **vbargs)
  File "C:\Users\Mathieu\Documents\pyvenv\mne\lib\site-packages\pyqtgraph\graphicsItems\ViewBox\ViewBox.py", line 407, in addItem
    if item.zValue() < self.zValue():
RuntimeError: wrapped C/C++ object of type AnnotRegion has been deleted
```

It occurs when I scroll by selecting the box on the overview, when I cross an annotation edge.

https://user-images.githubusercontent.com/73893616/160923886-b720b5cd-3ab1-479f-97c1-b0070add865d.mp4

Weirdly enough, if I now close the plotting window and do again in the same console `raw.plot()`, the new browser works as intended and this bug is not present. And second weird point, this bug is not present if I do the same 3 commands in an IPython console. :confused:

-----------

Reproducible example with MNE dataset:

```
import mne
fname = mne.datasets.sample.data_path() / 'MEG' / 'sample' / 'sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(fname, preload=True)
annotations = mne.Annotations([1, 10, 20, 30], [1, 2, 1, 2], ['a', 'b', 'a', 'b'])
raw.plot()
```

marsipu:
#### Reference issue
Fixes #82

This fixes an issue where in an interactive python console some Qt-Objects had been deleted if they hadn't been added to the Scene.
Before on every change of X-Range, Qt-Objects for Annotations and Events had been added and removed according to if they appear in this X-Range or not. In early testing this had been proven to be more performant with large sets of Annotations/Events.
This PR removes that and always adds all Qt-Objects for Annotations and Events to the Scene where Qt handles everything then.
With reasonable amounts of Annotations/Events (probably most common usecase) this new solutionis probably as good if not even better.
IMHO running the browser in an interactive python console without issue outweighs a potential benefit the former method had in some rarer cases.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1233,8 +1233,6 @@ class RawViewBox(ViewBox):
                                                     description=description,
                                                     values=(self._drag_start,
                                                             drag_stop))
-                    self.mne.plt.addItem(self._drag_region)
-                    self.mne.plt.addItem(self._drag_region.label_item)
                 elif event.isFinish():
                     drag_stop = self.mapSceneToView(event.scenePos()).x()
                     self._drag_region.setRegion((self._drag_start, drag_stop))
@@ -1368,14 +1366,17 @@ def _q_font(point_size, bold=False):
 class EventLine(InfiniteLine):
     """Displays Events inside Trace-Plot"""
 
-    def __init__(self, pos, label, color):
+    def __init__(self, mne, pos, label, color):
         super().__init__(pos, pen=color, movable=False,
                          label=str(label), labelOpts={'position': 0.98,
                                                       'color': color,
                                                       'anchors': [(0, 0.5),
                                                                   (0, 0.5)]})
+        self.mne = mne
         self.label.setFont(_q_font(10, bold=True))
         self.setZValue(0)
+
+        self.mne.plt.addItem(self)
 
 
 class Crosshair(InfiniteLine):
@@ -1990,6 +1991,9 @@ class AnnotRegion(LinearRegionItem):
         self.sigRegionChanged.connect(self.update_label_pos)
 
         self.update_color()
+
+        self.mne.plt.addItem(self)
+        self.mne.plt.addItem(self.label_item)
 
     def _region_changed(self):
         self.regionChangeFinished.emit(self)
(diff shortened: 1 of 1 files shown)


marsipu:
@larsoner Are the remaining failing CIs related to some xvfb or similar problems on Ubuntu for older mne-versions?

mne_qt_browser/_pg_figure.py
```diff
@@ -1368,15 +1366,18 @@ def _q_font(point_size, bold=False):
...
     """Displays Events inside Trace-Plot"""
 
-    def __init__(self, pos, label, color):
+    def __init__(self, mne, pos, label, color):
         super().__init__(pos, pen=color, movable=False,
                          label=str(label), labelOpts={'position': 0.98,
                                                       'color': color,
                                                       'anchors': [(0, 0.5),
                                                                   (0, 0.5)]})
+        self.mne = mne
```
hoechenberger, line 1375: You sure this should become part of the public API?

hoechenberger:
Nice, much less code than before!!!!

mne_qt_browser/_pg_figure.py
marsipu: EventLine shouldn't really used by the users. All other Qt-Objects also have `self.mne`, I just added it for consistency here in the same way. I think this is something originating from mne-python too where `mne` is also a public attribute of the mpl-browser.

marsipu:
> @larsoner Are the remaining failing CIs related to some xvfb or similar problems on Ubuntu for older mne-versions?

I think I solved that one in #140, never mind

mne_qt_browser/_pg_figure.py
hoechenberger: Ok!

hoechenberger (approved):

larsoner:
FYI I added `main` branch protection rules so auto-merge-when-green should be doable now

PR #140, opened 2022-08-14
BUG: Always newest xvfb-fix from mne

marsipu:

This fixes failing maintenance-CIs for backwards-compatitibility. With mne<1.1, the CIs on `ubuntu-latest` fail, because a fix from mne-tools/mne-python#10923 hadn't been implemented then.

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -100,7 +100,9 @@ jobs:
           powershell gl-ci-helpers/appveyor/install_opengl.ps1
         name: Setup OpenGL on Windows
         if: runner.os == 'Windows'
-      - run: ./tools/setup_xvfb.sh
+      - run: |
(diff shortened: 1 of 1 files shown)


hoechenberger (approved):
LGTM

@marsipu We could enable protected branches & automerge on CI pass for this repository, WDYT?

marsipu:
> @marsipu We could enable protected branches & automerge on CI pass for this repository, WDYT?

Why not, do you have the same over at `mne-python`?

hoechenberger:
> > @marsipu We could enable protected branches & automerge on CI pass for this repository, WDYT?
> 
> 
> 
> Why not, do you have the same over at `mne-python`?

We used to, not sure if it's still active. But I just recently enabled it for MNE-BIDS and the BIDS Pipeline.

larsoner:
Yes it's active there and I just activated it here

PR #145, opened 2022-09-26
BUG: Fix ref bug using weakref

larsoner:
In https://github.com/mne-tools/mne-python/pull/11205 you can see reference problems pop up on CircleCI, GH Actions, and Azure when we assert that garbage can be collected properly on plotter close. This is because PyQt5/PyQt6 appear to have issues when `signal.connect` is called using a `partial` of a bound method -- `gc` can't seem to resolve these on its own. I think we can fix them using `weakref.WeakMethod`, which I've done here.

Also becomes more selective about error resolution by only allowing disconnect errors when it's a separator (i.e., `action.text() == ''`, since `addSeparator` actually creates a `QAction` in the toolbar).

To verify this is the correct fix, though, we should try using this branch in https://github.com/mne-tools/mne-python/pull/11205 and make sure CIs come back green.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -11,11 +11,11 @@ import gc
 import math
 import platform
 import sys
+import weakref
 from pathlib import Path
 from ast import literal_eval
 from collections import OrderedDict
 from copy import copy
-from weakref import WeakMethod
 from os.path import getsize
 
 import numpy as np
@@ -169,8 +169,9 @@ class DataTrace(PlotCurveItem):
 
     def __init__(self, main, ch_idx, child_idx=None, parent_trace=None):
         super().__init__()
-        self.main = main
+        self.weakmain = weakref.ref(main)
         self.mne = main.mne
+        del main
 
         # Set clickable with small area around trace to make clicking easier.
         self.setClickable(True, 12)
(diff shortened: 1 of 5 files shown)


mne_qt_browser/_pg_figure.py
```diff
@@ -3137,13 +3158,15 @@ def _hidpi_mkPen(*args, **kwargs):
             'home': {
                 'alias': dur_keys[0],
                 'qt_key': Qt.Key_Home,
+                'kw': 'step',
```
larsoner, line 3161: This annoying thing is because the `.trigger()` in the fake mouse click function seems to cause the callback to be called with the signature `triggered(isChecked)`

larsoner:
Okay ea31499 is green running the improved MNE-Python tests from mne-tools/mne-python#11205, I'm going to revert the use of that branch. Then once that comes back green, I think we should merge and cut a new release!

larsoner:
Looks like I spoke too soon before. To make PySide6 happy I had to change a bunch of `self.main = main` to `self.weakmain = weakref.ref(main)` in the child dialog classes. I also refactored the speed tests to use a proper class rather than global variables, which I think is a cleaner (and more DRY) design for something that's meant to preserve and update states.

Tests passed with my more complete MNE-Python tests (from my `mne-python check` branch) on 9ffadf0, so I'll push a commit to revert using that branch in this PR, then hopefully we can merge and release.

marsipu:
Thank you very much @larsonergood to know that gc has

marsipu (approved):

PR #157, opened 2023-03-06
Change req. from sklearn to scikit-learn

mscheltienne:
Unless I am mistaken, `pip install sklearn` does not work anymore; in favor of `pip install scikit-learn`.


larsoner:
Cherry-picked into #158 with CI fixes, will give co-author credit, thanks @mscheltienne !

PR #158, opened 2023-03-06
MAINT: Fix CIs

larsoner:
Unclear why https://github.com/mne-tools/mne-qt-browser/pull/157 isn't running tests, but our most recent CI run (a month ago!) failed, so let's try this.

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -13,9 +13,9 @@ jobs:
   flake:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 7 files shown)


larsoner:
The trick was to go to the specific workflow in https://github.com/mne-tools/mne-qt-browser/actions/workflows/qt_viz_tests.yml and click "reenable" because it had been disabled due to inactivity. I'll mark this for merge-when-green since the last scheduled runs were red

larsoner:
I bumped the min MNE req to 1.0 here since it's about a year old now

PR #162, opened 2023-03-30
ENH: Speed up

### LINKED ISSUE #161: Scrolling (and general) performance with large number of annotations
With the current version of the browser, a large number of annotations make the browser very sluggish, or even temporarily freezes it. A couple of times today, I had the lovely pop-up: Force to Quit or Wait?

A couple of applications with large number of annotations that I encountered in the past months: 
- sleep recording, 7/8 hours with annotations every 30s to denote e.g. the sleep stage
- 20' recording, 64 channels EEG, with a video game based task where the button press (mouse and keyboard) are logged and saved as annotations

As you can imagine, especially with the second one, the number of annotations sky rocks: 2 876 segments in 20'.

At the moment, I have no idea how the annotations are handled by the browser and what is causing the bottleneck, but it would be good to discuss improvements and optimization of the annotation rendering. After all, the browser does not display that many annotation at once in the view area.

larsoner:
@mscheltienne can you see if being on this branch and https://github.com/mne-tools/mne-python/pull/11614 helps your use case at all? For me 3000 annotations is pretty usable but 10000 gets pretty slow still.

Closes #161

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -169,7 +169,13 @@ jobs:
           python -m pip install --upgrade pip
           git clone -b ${MNE_BRANCH} --single-branch --branch main https://github.com/mne-tools/mne-python.git ../mne-python
           python -m pip install -qe ../mne-python
-          python -m pip install -ve .${PIP_OPTION} -r requirements.txt -r requirements_testing.txt $QT_LIB
+          # Avoid a problematic version of PySide6
+          if [[ "$QT_LIB" == "PySide6" ]]; then
+            QT_LIB_INSTALL="PySide6!=6.5"
+          else
+            QT_LIB_INSTALL="$QT_LIB"
+          fi
+          python -m pip install -ve .${PIP_OPTION} -r requirements.txt -r requirements_testing.txt $QT_LIB_INSTALL
       - shell: bash -el {0}
         run: ./tools/get_testing_version.sh
         name: 'Get testing version'
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -193,6 +193,8 @@ jobs:
       - run: sudo apt install -yq libegl1
         name: Install EGL
         if: runner.os == 'Linux'
+      - name: Start Qt on its own
+        run: LD_DEBUG=libs python -c "from ${QT_LIB}.QtWidgets import QApplication, QWidget; app = QApplication([]); import matplotlib; matplotlib.use('QtAgg'); import matplotlib.pyplot as plt; plt.figure()"
       - name: Check Qt
         run: |
           mne sys_info -pd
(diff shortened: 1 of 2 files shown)


larsoner:
@mscheltienne assuming this snippet:
```
import numpy as np
import mne
raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() / 'MEG' / 'sample' / 'sample_audvis_raw.fif')
raw.pick('eeg').load_data()
raw = mne.concatenate_raws([raw.copy() for _ in range(int(20 * 60 / raw.times[-1]))])  # -> ~20 min
onsets = np.linspace(0, raw.times[-1] - 0.2, 3000, endpoint=False)
raw.set_annotations(mne.Annotations(onsets, 0.1, 'TEST'))
raw.plot()
```
replicates your problem, let's work from that so we're comparing apples to apples. If it doesn't replicate your problem, can you modify it a bit until it does?

For me on `main` of both mne-qt-browser and mne-python I get ~18sec to it being usable:

https://user-images.githubusercontent.com/2365790/229546580-26ac2e7d-f5c2-4b83-9d51-8382004f637b.mp4

And on my two `speed` branches I get closer to ~12 sec:

https://user-images.githubusercontent.com/2365790/229546612-785b005f-d2d6-4782-83e7-168ad24c21d2.mp4

So it should at least help somewhat at the loading end.

For the actual scrolling I think we'll need to `setVisible(False)` by default and use NumPy vectorized arithmetic to decide which to `setVisible(True)` to based on the view window. That way Qt/OpenGL/whatever doesn't have to decide what's visible based on viewbox transforms (or whatever it's doing currently). I'll work on that now...

larsoner:
Okay now scrolling is much faster, too:


https://user-images.githubusercontent.com/2365790/229554267-b228b792-989c-4073-8ef9-ce3b9a8a6fd3.mp4

mscheltienne:
Yes that looks way better already!

mscheltienne:
Also it looks like one of those 2 PR is affecting the alpha on the region, the colors are more vivid/less transparent and I'm loosing the labels at the bottom of each region (probably hidden by the lack of transparency?).

mscheltienne:
The changes LGTM. One introduced bug however: deleting an annotation does not remove its label anymore. It seems this is now evaluated to False: https://github.com/larsoner/mne-qt-browser/blob/92d607c7cf99c98103177a7ea8e3ecf3fe43b766/mne_qt_browser/_pg_figure.py#L4008

```
import mne


raw = mne.io.read_raw_fif(
    mne.datasets.sample.data_path()
    / "MEG"
    / "sample"
    / "sample_audvis_raw.fif"
)
raw.pick("eeg").load_data()

raw = mne.concatenate_raws(
    [raw.copy() for _ in range(int(20 * 60 / raw.times[-1]))]
)  # -> ~20 min
raw.plot()
```

[Screencast from 04-04-2023 10:07:10 AM.webm](https://user-images.githubusercontent.com/73893616/229728813-9ba3d570-aab8-40d5-ab5f-426949f608b5.webm)

larsoner:
Thanks for the hint, works locally for me now @mscheltienne . Can you double check? Then feel free to merge if you're happy :)

mscheltienne:
Works like a charm, thanks for working on this @larsoner!

cbrnr:
Great work @larsoner, thanks! This should be worth a new release, no?

larsoner:
https://pypi.org/project/mne-qt-browser/0.5.0/

PR #166, opened 2023-06-05
MAINT: Improve PySide6 compat

larsoner:
PySide6 version 6.5 should in principle be supported. I expect this to fail, hopefully the fixes don't end up being too bad...


larsoner:
All green and this is just a CI change so no need for a point release I think. TL;DR is that `PySide6!=6.5.1` should work now (e.g., including 6.5.0)

PR #168, opened 2023-06-09
Fix scrolling through data with a visualization filter

### LINKED ISSUE #167: Filter, scrolling, whitening are not playing nicely together
A bug reported today by  @ArthurNguyen on our site, present when running the code snippet below on mne-python/mne-qt-browser `main`. The bug is absent from the `matplotlib` backend.

```
from mne.viz import set_browser_backend


set_browser_backend("qt")
directory = sample.data_path() / "MEG" / "sample"
raw = read_raw_fif(directory / "sample_audvis_raw.fif", preload=False)
raw.pick_types(eeg=True)
raw.load_data()
raw.plot(theme="light", highpass=1., lowpass=10.)
```

[Screencast from 06-08-2023 05:07:25 PM.webm](https://github.com/mne-tools/mne-qt-browser/assets/73893616/e52ab43f-a020-4439-bac5-13a1628e0f3e)

Scrolling and/or clicking on the overview bar seems to disables whitening (or whatever is done by `w`). I'll try to have a look tomorrow.

mscheltienne:
Fixes #167, `mne.global_data` was not correctly pre-computed because of an indexing error. 

~~It looks to me like `main._rerun_precompute()` must be run before updating the xRange, to run through https://github.com/mne-tools/mne-python/blob/d158117a716512fbe96d8ff6c4d735f1e9601d09/mne/viz/_figure.py#L349-L385 and to apply the visualization filters.~~

--------------

~~For the click on the overview bar, I'm lacking a good idea to fix it. The update happens in the OverviewBar class, here: https://github.com/mne-tools/mne-qt-browser/blob/f07ab8a2509b6d76c1fca9d2a2326c545a25eb1c/mne_qt_browser/_pg_figure.py#L1050~~
~~and it doesn't have a reference to `main`, to `_rerun_precompute` or to `load_thread`..~~

--------------

~~There is also this update in the `TimeScrollBar` class: https://github.com/mne-tools/mne-qt-browser/blob/f07ab8a2509b6d76c1fca9d2a2326c545a25eb1c/mne_qt_browser/_pg_figure.py#L693-L701
~~and I have no idea when it is triggered with `self.external_change` set to False.~~

--------------

And as a side-change, I added `self.mne.noise_cov is not None` to disable `w`-key if there is nothing to do.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3444,6 +3444,9 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             xmax = self.mne.xmax
             xmin = xmax - self.mne.duration
 
+        if self.mne.filter_coefs is not None:
+            self._rerun_precompute()
+
         self.mne.plt.setXRange(xmin, xmax, padding=0)
 
     def vscroll(self, step):
@@ -3503,6 +3506,9 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
 
         if xmin < 0:
             xmin = 0
+
+        if self.mne.filter_coefs is not None:
+            self._rerun_precompute()
 
         self.mne.ax_hscroll.update_duration()
         self.mne.plt.setXRange(xmin, xmax, padding=0)
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2606,7 +2606,7 @@ class LoadThread(QThread):
         # Deactive remove dc because it will be removed for visible range
         stashed_remove_dc = self.mne.remove_dc
         self.mne.remove_dc = False
-        data = browser._process_data(data, 0, len(data), picks, self)
+        data = browser._process_data(data, 0, data.shape[-1], picks, self)
         self.mne.remove_dc = stashed_remove_dc
 
         self.mne.global_data = data
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -4174,10 +4174,11 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             self.mne.fig_proj.toggle_all()
 
     def _toggle_whitening(self):
-        super()._toggle_whitening()
-        # If data was precomputed it needs to be precomputed again.
-        self._rerun_precompute()
-        self._redraw()
+        if self.mne.noise_cov is not None:
+            super()._toggle_whitening()
+            # If data was precomputed it needs to be precomputed again.
+            self._rerun_precompute()
+            self._redraw()
 
     def _toggle_settings_fig(self):
         if self.mne.fig_settings is None:
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3444,9 +3444,6 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             xmax = self.mne.xmax
             xmin = xmax - self.mne.duration
 
-        if self.mne.filter_coefs is not None:
-            self._rerun_precompute()
-
         self.mne.plt.setXRange(xmin, xmax, padding=0)
 
     def vscroll(self, step):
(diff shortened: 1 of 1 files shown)


larsoner:
> and it doesn't have a reference to main, to _rerun_precompute or to load_thread..

You can add a ref to `main` / `self.mne` or whatever if you need to. You might need to make it a `weakref` though just to make sure we don't get a bad reference cycle.

> It looks to me like main._rerun_precompute() must be run before updating the xRange

Based on what it looks like in your changes, `_rerun_precompute()` gets called inside `hscroll` which is strange. I'd expect `_rerun_precompute()` to precompute everything needed for display (maybe including filtering etc.). So it seems like you're potentially recomputing stuff for the entire duration every time a horizontal scroll/move is done, is that right? If not, can you add some comments inline in the code to make it clear what's actually happening? Like if it's just recomputing for the given segment or something...

mne_qt_browser/_pg_figure.py
```diff
@@ -3444,6 +3444,9 @@ def hscroll(self, step):
             xmax = self.mne.xmax
             xmin = xmax - self.mne.duration
 
+        if self.mne.filter_coefs is not None:
+            self._rerun_precompute()
```
larsoner, line 3448: This seems like a weird conditional to me. Why does it matter if there are filter coefs or not that you have to recompute just for a scroll event?

mne_qt_browser/_pg_figure.py
```diff
@@ -3504,6 +3507,9 @@ def change_duration(self, checked=False, *, step):
         if xmin < 0:
             xmin = 0
 
+        if self.mne.filter_coefs is not None:
+            self._rerun_precompute()
```
larsoner, line 3511: Same -- this seems like the wrong place to need to rerun precomputation. Shouldn't this just be run when how the signal is processed (e.g., projectors, whitening, etc.) is changed rather than being done inside a duration change event?

mne_qt_browser/_pg_figure.py
```diff
@@ -4168,10 +4174,11 @@ def _toggle_all_projs(self):
...
 
     def _toggle_whitening(self):
-        super()._toggle_whitening()
-        # If data was precomputed it needs to be precomputed again.
-        self._rerun_precompute()
-        self._redraw()
+        if self.mne.noise_cov is not None:
+            super()._toggle_whitening()
+            # If data was precomputed it needs to be precomputed again.
+            self._rerun_precompute()
```
larsoner, line 4180: This one *does* make sense to me. The other two I'm confused by...

mscheltienne:
> You can add a ref to main / self.mne or whatever if you need to.

There is already a weakref to main, but when I used it, the overview bar froze 😞 

> So it seems like you're potentially recomputing stuff for the entire duration every time a horizontal scroll/move is done, is that right?

That's indeed what I added to this PR. It does fix the issue, but it's indeed very weird, and it's probably not the right fix. I'll have to dig more into this one.

mscheltienne:
Quick update:
- For `self._rerun_precompute()`, you are completely right. I thought the browser was pulling and processing data view by view, and the `start` and `stop` arguments of `_process_data` reinforced this idea. Setting `precompute=False` "fixes" the issue, by forcing to recompute each view as I did in this PR.
- The bug shows different data traces compared to unfiltered raw because of the decimation with a factor of 20. If those lines are skipped https://github.com/mne-tools/mne-qt-browser/blob/f07ab8a2509b6d76c1fca9d2a2326c545a25eb1c/mne_qt_browser/_pg_figure.py#L355-L358 scrolling passed the first view removes the filter, and we get back the unfiltered `raw` instead of an in-between.

Still looking for the root cause..

mscheltienne:
One more weird point to investigate.. 

```
from matplotlib import pyplot as plt
from mne.viz import set_browser_backend


set_browser_backend("qt")
directory = sample.data_path() / "MEG" / "sample"
raw = read_raw_fif(directory / "sample_audvis_raw.fif", preload=False)
raw.pick("EEG 001")
browser = raw.plot(theme="light", highpass=1., lowpass=10., precompute=True)
```

This code will go 4 times through https://github.com/mne-tools/mne-python/blob/c98749b529a41473cfccbc18485a9d1924546545/mne/viz/_figure.py#L326-L347 with the following values for `starts` and `stops` variables:

```
[0] [6006]
[0] [6006]
[0] [6006]
[0] [1]
```

I would expect this function to be called only once, on the entire array or size 166800.

mne_qt_browser/_pg_figure.py
```diff
@@ -2606,7 +2606,7 @@ def run(self):
         # Deactive remove dc because it will be removed for visible range
         stashed_remove_dc = self.mne.remove_dc
         self.mne.remove_dc = False
-        data = browser._process_data(data, 0, len(data), picks, self)
+        data = browser._process_data(data, 0, data.shape[-1], picks, self)
```
larsoner, line 2609: 😬

larsoner:
Thanks @mscheltienne ! Worth a point release. But first I'd like to investigate an arrow key bug I noticed on some Qt6 variant. I'll try to get to it today.

PR #169, opened 2023-06-12
BUG: Fix bug with arrow-key-based scrolling

larsoner:
I found that the left/right arrow keys were scrolling by 3-4x the current `duration` when they should have scrolled by 1/4 of `duration`. This was because `scroll_sensitivity` was being included when a keypress was used. This fixes it.


larsoner:
Actually I think I got the epoch step wrong, will fix then merge

PR #171, opened 2023-06-21
Lower zvalue from bad traces to 0

### LINKED ISSUE #170: [ENH] Make it possible to toggle visibility of bad channels, especially in butterfly mode
I like a lot the butterfly mode, which helps to find big issues in your recording. But it's not that handy on recordings with too many bad channels:

Screenshot from 2023-06-21 14-57-25

It would be nice if we could toggle off the visibility of the bad traces.

mscheltienne:
Default z value for a trace is 1. This PR lowers it to 0 for bad traces which are not as important.
Partially fix #170

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -417,6 +417,10 @@ class DataTrace(PlotCurveItem):
 
             # Update line color status
             self.isbad = not self.isbad
+            if self.isbad:
+                self.setZValue(0)
+            else:
+                self.setZValue(1)
 
             # Update colors for epochs
             if self.mne.is_epochs:


larsoner:
I don't think the fix is partial but rather complete -- zero-alpha channels are already properly supported by the `bad_color` arg! But if you want some interactive toggle then I guess it's partial, and we should add a color picker sort of thing somewhere

mne_qt_browser/_pg_figure.py
```diff
@@ -417,6 +417,10 @@ def toggle_bad(self, x=None):
 
             # Update line color status
             self.isbad = not self.isbad
+            if self.isbad:
+                self.setZValue(0)
+            else:
+                self.setZValue(1)
```
larsoner, line 423: Looks like this will only work with an interactive toggle. Does it already work properly for channels that are in `info['bads']` on initial plot?

mscheltienne: Arf, we are not running through that method on init as I thought..

mscheltienne: Next commit does the trick, but it feels a bit ugly.. let me check if I can improve this somehow.

larsoner: Rather than code dup naively I would have expected it to be in `self.update_color()` because it looks like that's called by this method, and I'd assume called in `__init__` as well. Did you think about/consider this?

drammock (approved):
IMO this is enough; personally I'd stop short of adding a "hide bads" toggle and/or a color picker.

larsoner:
Thanks @mscheltienne !

PR #172, opened 2023-06-26
Use weakmain in the annotation dialog box

mscheltienne:
A very unreliable bug reported today at our site, I managed to observe it twice on 2 different computers, among which mine but I am unable to reliably reproduce it. After playing with annotations, upon closing the browser, the following error is raised:

```
File "[...]/mne_qt_browser/_pg_figure.py", line 2305, in _edit_description_dlg
    _AnnotEditDialog(self)
File "[...]/mne_qt_browser/_pg_figure.py", line 2102 in __init__
    super().__init__(annot_dock.main, title=[...])
AttributeError: 'AnnotationDock' object has no attribute 'main'
```

Yes, I had to write the traceback from a screenshot :sob: 
My guess is that `annot_dock.main` is garbage collected early. Changing to a weakref does not seem to break anything, and that might fix this issue?

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2099,7 +2099,7 @@ class AnnotRegion(LinearRegionItem):
 
 class _AnnotEditDialog(_BaseDialog):
     def __init__(self, annot_dock):
-        super().__init__(annot_dock.main, title='Edit Annotations')
+        super().__init__(annot_dock.weakmain, title='Edit Annotations')
         self.ad = annot_dock
 
         self.current_mode = None


mne_qt_browser/_pg_figure.py
```diff
@@ -2099,7 +2099,7 @@ def update_label_pos(self):
 
 class _AnnotEditDialog(_BaseDialog):
     def __init__(self, annot_dock):
-        super().__init__(annot_dock.main, title='Edit Annotations')
+        super().__init__(annot_dock.weakmain, title='Edit Annotations')
```
larsoner, line 2102: I would expect it to need to be this to resolve to the object itself or to None
```suggestion
        super().__init__(annot_dock.weakmain(), title='Edit Annotations')
```

mscheltienne: :see_no_evil: Yes.. without it breaks the annotation edit.

mscheltienne:
I got the error a third time, the weakref did not help..

larsoner:
> I got the error a third time, the weakref did not help..

But this time the error must have been `AttributeError: 'AnnotationDock' object has no attribute 'weakmain'`, right?

One option is just to do `getattr(annotdock, 'main', None)` or `getattr(annotdock, 'weakmain', lambda: None)()`. You might also need to short-circuit somewhere else later.

But I would maybe go to `File "[...]/mne_qt_browser/_pg_figure.py", line 2305, in _edit_description_dlg` and only initialize if `hasattr(self, 'main')` or so.

Maybe you are hitting https://bugreports.qt.io/browse/PYSIDE-2299 for example on some PySide 6.5 variant. Personally I only use PyQt6 at the moment because PySide6 still does not set the inputhook properly for REPL use with matplotlib interactive mode :(

mscheltienne:
342876871dfb5cc063498773159213a89e616cdf prevents creating annotations outside of `t=(0, raw.times[-1])`. Before, it was possible to create an annotation with a `drag_start` or `drag_stop` outside of those boundaries which doesn't make a lot of sense and would break the overview bar.

At the moment, it's still possible to drag an existing annotation outside of `t=(0, raw.times[-1])`. IMO, as for the overview bar, it should block drag outside of those boundaries. I haven't found yet which part of the code is responsible for that.

Also, not sure how this works exactly: https://github.com/mne-tools/mne-qt-browser/blob/cc41701f4ca4677456b1b201d4682fc7324fba51/mne_qt_browser/_pg_figure.py#L2086-L2087
But IMO, this should be a bug instead of the desired behavior. When having overlapping annotations, deleting becomes very difficult as every annotation overlapping is deleted instead of the selected (and highlighted) annotation.

-----------

Back to the initial issue. the `_AnnotEditDialog` box to edit annotation name. I have **no idea** how the error above is raised sometimes on `main`, I actually got it once more on `main`, *without* clicking on the `Edit description` which initialize this object.. But regardless, `Edit description` is broken on main and doesn't open the pop-up at all. At least, when using this PR and `weakmain()` the pop-up does open.

And finally, now that it opens, you can try using it :sweat: And it fails. This 2 functions are broken: https://github.com/mne-tools/mne-qt-browser/blob/cc41701f4ca4677456b1b201d4682fc7324fba51/mne_qt_browser/_pg_figure.py#L2253-L2301
At least the traceback in the console issued when using the annotation edition tool is useful!

-----------

Working on it slowly when I have the time ;)

mne_qt_browser/_pg_figure.py
```diff
@@ -2209,13 +2216,19 @@ def _init_ui(self):
...
         layout.addWidget(self.start_bx)
 
         layout.addWidget(QLabel('Stop:'))
         self.stop_bx = QDoubleSpinBox()
         self.stop_bx.setDecimals(time_decimals)
-        self.stop_bx.editingFinished.connect(self._stop_changed)
+        self.stop_bx.setMinimum(1 / self.mne.info["sfreq"])
+        self.stop_bx.setMaximum(self.mne.inst.times[-1])
+        self.stop_bx.setSingleStep(0.05)
+        self.stop_bx.valueChanged.connect(self._stop_changed)
```
mscheltienne, line 2231: The spinBox were free to do whatever. Now they are bound to sensible min/max/step; and the selected region updates on `valueChanged` instead of on `editingFinised` which gives better feedback to the user.

Also, the spinBox are disabled/reset when the *selected* annotation is removed and re-enabled on an annotation selection.

mscheltienne:
Still looking for the piece of code responsible for dragging an existing annotation around, and then that PR should be good to go.

hoechenberger (approved):
Very nice!!

mne_qt_browser/_pg_figure.py
```diff
@@ -4735,3 +4762,19 @@ def _init_browser(**kwargs):
...
+        if hasattr(super(), "__enter__"):
+            super().__enter__()
+        else:
+            super().reblock()
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if hasattr(super(), "__exit__"):
+            super().__exit__(exc_type, exc_value, traceback)
+        else:
+            super().unblock()
```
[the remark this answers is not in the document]
larsoner: Seems okay to me, but I'm not very familiar with this QSignalBlocker

larsoner:
Good to go @mscheltienne ? I assume it's not trivial to test this :(

mscheltienne:
To be honest, I have no idea how to write good test for Qt-related code. 
From all the manual tests I've been doing, seems to work fine, at least on PyQt5 and on PySide2.

mscheltienne:
If you want, I can however do a short screencast demonstrating the changes between main and this PR. It's easier than to write good tests 😅

larsoner:
We have some `_fake_click` functions and you can also use the `qtbot` fixture of pytest-qt. But we can save that for another day when you're motivated to learn it :)

Thanks @mscheltienne !

PR #173, opened 2023-06-27
Improve deletion of annotation

mscheltienne:
One more :)

With this one, I'd like to introduce a change in behavior when deleting an annotation. Before, all annotation overlapping the `Qt.RightButton` were deleted. This is inconvenient because only one annotation would be highlighted and yet all annotations clicked would get deleted.
With this change, only the highlighted annotation, on top of the others, is actually deleted.

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, windows, macos]
-        mne: [main, maint/1.4]
+        mne: [annot-deletion, maint/1.4]
         opengl: ['opengl']
         python: ['3.11']
         name: [matrix]
@@ -77,7 +77,7 @@ jobs:
         run: |  # use MNE main to ensure test files are available
           set -e
           python -m pip install --upgrade pip
-          git clone --single-branch --branch annot-deletion https://github.com/drammock/mne-python.git ../mne-python
+          git clone -b ${MNE_BRANCH} --single-branch https://github.com/drammock/mne-python.git ../mne-python
           python -m pip install -qe ../mne-python
           if [[ "${{ matrix.opengl }}" == "opengl" ]]; then
             PIP_OPTION="[opengl,tests]"
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -77,7 +77,7 @@ jobs:
         run: |  # use MNE main to ensure test files are available
           set -e
           python -m pip install --upgrade pip
-          git clone -b ${MNE_BRANCH} --single-branch --branch main https://github.com/mne-tools/mne-python.git ../mne-python
+          git clone -b ${MNE_BRANCH} --single-branch --branch annot-deletion https://github.com/drammock/mne-python.git ../mne-python
           python -m pip install -qe ../mne-python
           if [[ "${{ matrix.opengl }}" == "opengl" ]]; then
             PIP_OPTION="[opengl,tests]"
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -77,7 +77,7 @@ jobs:
         run: |  # use MNE main to ensure test files are available
           set -e
           python -m pip install --upgrade pip
-          git clone -b ${MNE_BRANCH} --single-branch --branch annot-deletion https://github.com/drammock/mne-python.git ../mne-python
+          git clone --single-branch --branch annot-deletion https://github.com/drammock/mne-python.git ../mne-python
           python -m pip install -qe ../mne-python
           if [[ "${{ matrix.opengl }}" == "opengl" ]]; then
             PIP_OPTION="[opengl,tests]"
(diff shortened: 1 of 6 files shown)


mscheltienne:
It also helps in selecting annotation as the one currently set in the description combo-box on the top left will take priority.

mscheltienne:
I was not expecting to break all the tests.. the `matplotlib` backend has the same inconvenient behavior thus it would require the same change if everyone agrees it's reasonable.

larsoner:
@drammock probably a usage question for you above ☝️

drammock:
@mscheltienne do you have an MWE you're using to debug/test this? If not I can throw one together, then take a look at the matplotlib browser

mscheltienne:
Here you go:

```
from mne import Annotations
from mne.viz import set_browser_backend


set_browser_backend("qt")

fname = sample.data_path() / "MEG" / "sample" / "sample_audvis_raw.fif"
raw = read_raw_fif(fname, preload=False)
raw.crop(0, 60)
raw.pick("eeg")
raw.load_data()
annotations = Annotations(onset=[3, 3.4, 3.7], duration=[2, 1, 3], description=["1", "bad_2", "3"])
raw.plot(theme="light", highpass=1, lowpass=40)
```

On main, if you enable annotation mode and then right-click on the overlapping part, all annotations are deleted.
On the Qt browser, it's even worst than with the matplotlib backend because the highlighting tricks you in thinking that only that annotation is impacted.

drammock:
this is not going to be straightforward for matplotlib browser, as all the annotations end up with the same z-order 😞 so either we need to:

- refactor the z-order handling so each annotation has a distinct z-order
- manually keep track of the order in which spans are created and prefer to delete the most recent one
- use some other criterion (like order of the descriptions in the annotations list?)

Any thoughts on best approach here?

mscheltienne:
I don't know well the matplotlib backend, but refactoring the z-order seems logic.

> use some other criterion (like order of the descriptions in the annotations list?)

I would say: give the currently selected description priority.

larsoner:
This version number is certainly weird!

https://github.com/mne-tools/mne-qt-browser/actions/runs/5660718414/job/15337067555?pr=173#step:10:26

```
├☑ mne               0.10.0.dev9408+gd70215262
```

larsoner:
And apparently the `#egg` made it so the test files weren't included

larsoner:
Actually looks like you have the tags, maybe your `main` needs to be updated 🤷

drammock:
> Actually looks like you have the tags, maybe your `main` needs to be updated 🤷

I just pushed the `1.4.x` tags a few minutes ago (after your comment)

drammock:
ok on https://github.com/mne-tools/mne-qt-browser/pull/173/commits/f8056d69e98b4b60357f1306cffa92cf733d8142 this looks normal now:
```
 ├☑ mne               1.5.0.dev82+gd5c60e104
```
but the test files are still missing:

```
E   FileNotFoundError: fname does not exist: "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/mne/io/tests/data/test_raw.fif"
```

Looking at the `install dependencies` step, I see MNE getting installed twice:

- once with the `git clone --branch ...` line, here: https://github.com/mne-tools/mne-qt-browser/actions/runs/5661989593/job/15341031820?pr=173#step:4:38
- once from the pyproject.toml install spec, here: https://github.com/mne-tools/mne-qt-browser/actions/runs/5661989593/job/15341031820?pr=173#step:4:109

Then some warning lines about missing files when reading the manifest: https://github.com/mne-tools/mne-qt-browser/actions/runs/5661989593/job/15341031820?pr=173#step:4:273

then a series of warnings about `mne.channels.data.layouts`, `montages`, `fsaverage`, `helmets`, `image`, `_infant`, ... (several more): https://github.com/mne-tools/mne-qt-browser/actions/runs/5661989593/job/15341031820?pr=173#step:4:1164

larsoner:
I would get rid of the pyproject one, it's probably not installing the way we want

drammock:
OK!  All of the CI matrix jobs using `annot-deletion` branch are passing!  Which means (I think) that once https://github.com/mne-tools/mne-python/pull/11760 is in and the changes to CI config / pyproject.toml are reverted here, *all* the CIs here should pass. So in theory we can now:

1. revert the changes to CI config / pyproject.toml here
2. cross fingers
3. merge both PRs
4. profit!

But to be extra cautious, I'll push a temporary commit to https://github.com/mne-tools/mne-python/pull/11760 that uses this branch, that should hopefully get *all* upstream CIs green.

larsoner:
🤞

larsoner:
Actually the `maint/1.4` CI fails here on the last two commits so I think we'll have to work a bit harder on that. Basically we need those tests *not* to run on 1.4 (because behavior was wrong there and correct here) so we need to figure that out somehow. One not-too-difficult option would be to apply a patch in the 1.4 case that `@pytest.mark.xfail`s them.

drammock:
> Actually the `maint/1.4` CI fails here on the last two commits

yeah I think those should now `xfail` on `maint/1.4`. Seems easier than trying to triage versions and test for the old (arguably broken) behavior

PR #174, opened 2023-06-28
Remove shortcut "h" plotting epoch PTP amp when viewing raw

mscheltienne:
So, pressing "h" on a `Raw` would not do anything, until you close the browser and open it again. Nice traceback ;)

The key `h` calls `_toggle_epoch_histogram`, which is not checking if we are plotting raw or epochs; while the function `_toggle_epoch_histogramm` with 2 `m` at the end does the same stuff correctly and is unused. 
This PR fixes that typo and removes the shortcut `h` entirely for raw recordings.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -4332,10 +4332,6 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         self.mne.remove_dc = not self.mne.remove_dc
         self._redraw()
 
-    def _toggle_epoch_histogram(self):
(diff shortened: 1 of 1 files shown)


mne_qt_browser/_pg_figure.py
```diff
@@ -3393,6 +3393,9 @@ def __init__(self, **kwargs):
         if self.mne.is_epochs:
             # Disable time format toggling
             del self.mne.keyboard_shortcuts['t']
+        else:
+            # dissable histogram of epoch PTP amplitude
```
drammock, line 3397:
```suggestion
            # disable histogram of epoch PTP amplitude
```

drammock (approved):

PR #177, opened 2023-07-27
MAINT: Patch old MNE ver test

larsoner:
Hopefully green then merge the 0.5.2 release

### DIFF
--- a/.github/workflows/mne_1_4.patch
+++ b/.github/workflows/mne_1_4.patch
@@ -1,0 +1,20 @@
+diff --git a/mne/viz/tests/test_raw.py b/mne/viz/tests/test_raw.py
+index 5533b5870..56ff1e607 100644
+--- a/mne/viz/tests/test_raw.py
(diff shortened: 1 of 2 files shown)


drammock:
wild, I never would have thought to actually use `patch` in the CI workflow like that. 🎩

mscheltienne:
And I had no idea `patch` even existed!

larsoner:
I got the idea from needing to patch various feedstocks (dipy, mayavi, etc.) for conda-forge 🙄 

For future reference it was fairly easy -- I went to `mne-python`, checked out `maint/1.4`, made the changes I wanted, and created a patch as:
```
git diff > ../mne-qt-browser/.github/workflows/mne_1_4.patch
```
Then I reset my changes to `mne-python`, `cd`ed to the `mne-qt-browser` dir, and checked that the `patch ...` command used here put my changes back where they belonged (and it did... after iterating on the command a bit!).

drammock:
all green! 🎉

PR #178, opened 2023-07-29
Merge annotations which are subject to a move/shrink/extend yielding an overlap with other annotations of the same type

mscheltienne:
On both the `matplotlib` and `qt` backend, creating a new annotation which overlap partially or fully with another annotation of the same type will merge both annotations.
On the `qt` backend, contrary to the `matplotlib` backend, annotations can be moved (left click and drag on the region) and resized (left click and drag on a region edge). I appreciate a lot this feature, but it does not merge newly overlapping annotations, which can be annoying.

For now, the changes don't work, but it's a start 😄 My main issue is that I don't understand why the signal `sigRegionChangeFinished` is emitted while an annotation is created, as long as you don't let go of the left click.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1253,10 +1253,12 @@ class RawViewBox(ViewBox):
                             event.lastScenePos()).x()
                     self._drag_start = 0 if self._drag_start < 0 else self._drag_start
                     drag_stop = self.mapSceneToView(event.scenePos()).x()
-                    self._drag_region = AnnotRegion(self.mne,
-                                                    description=description,
-                                                    values=(self._drag_start,
-                                                            drag_stop))
+                    self._drag_region = AnnotRegion(
+                        self.mne,
+                        description=description,
+                        values=(self._drag_start, drag_stop),
+                        weakmain=self.weakmain,
+                    )
                 elif event.isFinish():
                     drag_stop = self.mapSceneToView(event.scenePos()).x()
                     drag_stop = 0 if drag_stop < 0 else drag_stop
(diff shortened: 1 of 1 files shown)


drammock:
> On the `qt` backend, contrary to the `matplotlib` backend, annotations can be moved (left click and drag on the region) and resized (left click and drag on a region edge).

actually the MPL backend does allow resizing (but not moving) annotation spans. And it does already work that if you resize to the point where it overlaps another of the same type, they will merge. See the "draggable edges?" checkbox in its annotation dialog window.

drammock:
I guess to make them more consistent we should change the MPL checkbox to just say "draggable?" and allow dragging the whole span. It didn't ever occur to me that this would be useful, but you say it is so we should support it.

mscheltienne:
> See the "draggable edges?" checkbox 

I did not notice that checkbox! So indeed the matplotlib backend could be improved by changing the checkbox to `draggable` and by adding the feature. I would argue it could be checked by default, but that's only my opinion. 

And now the Qt backend behavior, not merging annotations, can be labelled as a bug.

mscheltienne:
Although, I don't think it's worth to spend time adding new annotation-related features to the `matplotlib` backend (or interactive features alltogether). Annotating is rather slow and difficult with that backend while it's simple and quick on the Qt one. IMO, the matplotlib backend is useful to create plots programmatically while the Qt backend is better for any interactive feature.

marsipu:
@mscheltienne Thank you for the suggestion, I like the idea too.
I was just thinking maybe in some cases you might want to have overlapping annotations and then it would also be a problem if we always merge when you move overlapping annotations.
Maybe something like holding `Shift` while moving could solve that?

mscheltienne:
I don't see a use case for interactive drawing of overlapping annotations with the _same_ description. Maybe for programmatically added annotation, and even then I can not come up with an example. Would you have one?

marsipu:
>I don't see a use case for interactive drawing of overlapping annotations with the same description. 

Oh of course, I didn't see that in the title. For the same type/description I don't have a usecase either

mscheltienne:
OK, that one should work. Let's see if the CIs are green.
In the end, I handled the channel merging directly in `mne-qt-browser` with the existing `_merge_annotations` function from `mne`, thus this PR is not intertwined with https://github.com/mne-tools/mne-python/pull/11855 :tada:

mscheltienne:
Edited the CIs to run with https://github.com/mne-tools/mne-python/pull/11855
Which should pass here on MNE main and on MNE annotations.

mscheltienne:
CIs passing with https://github.com/mne-tools/mne-python/pull/11855
https://github.com/mne-tools/mne-qt-browser/actions/runs/5751792494/job/15591341617

The 2 jobs failing are due to the branch used by those jobs missing on my fork.

PR #182, opened 2023-08-01
Improve settings dialog

marsipu:
### What does this implement/fix?
This improves the settings dialog by:
- [x] Adding an antialiasing checkbox
- [ ] adding a toggle for theme (as mentioned in #83)
- [ ] <other suggestions of parameters to expose?>

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1710,14 +1710,6 @@ class _BaseDialog(QDialog):
                 self.mne.child_figs.remove(self)
         event.accept()
 
-    # If this widget gets activated (e.g., the user clicks away from the
-    # browser but then returns to it by clicking in a selection window),
-    # the main window should be raised as well
-    def event(self, event):
-        if event.type() == QEvent.WindowActivate:
-            _qt_raise_window(self.weakmain())
-        return super().event(event)
-
 
 class SettingsDialog(_BaseDialog):
     """Shows additional settings."""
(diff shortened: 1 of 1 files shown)


mne_qt_browser/_pg_figure.py
```diff
@@ -1574,14 +1574,6 @@ def closeEvent(self, event):
...
         event.accept()
 
-    # If this widget gets activated (e.g., the user clicks away from the
-    # browser but then returns to it by clicking in a selection window),
-    # the main window should be raised as well
-    def event(self, event):
-        if event.type() == QEvent.WindowActivate:
-            _qt_raise_window(self.weakmain())
-        return super().event(event)
-
```
marsipu, line 1584: @larsoner I can't reproduce the problem you describe (at least not on windows). Does it still

https://github.com/mne-tools/mne-qt-browser/assets/46245704/70cf02d7-b44d-4f1a-b97d-72dd84608025

 happen for you?

The problem is, that this apparently prevents Checkboxes in the Dialog from working, because when they are clicked the focus is forced back to main and they can't trigger and change state.

larsoner: It used to happen to me on macOS but doesn't anymore. So I'd say let's remove it and if it turns out to be a problem we'll find a better fix

larsoner:
@marsipu should we just merge this as-is and proceed with follow-up enhancements for the other checkboxes later?

marsipu:
@larsoner yes, sorry forgot about this one.
Last thing I tried was switching while the app was running but I didn't succeed.
That could be subject of another PR together with adding the setting.
Could you unlink the dark-theme-issue before you merge so it doesn't get closed?
Thx

marsipu:
@larsoner yes, sorry forgot about this one.
Last thing I tried was switching while the app was running but I didn't succeed.
That could be subject of another PR together with adding the setting.
I am currently away from my PC but I can look at the merge conflicts next week

larsoner:
Marking for merge when green, thanks in advance @marsipu !

PR #183, opened 2023-08-02
Fix: remove description bug

marsipu:
## What does this fix?
This fixes an error when hitting "Remove description" when all are already removed

## How to reproduce the error:
```python
import mne
import os
import numpy as np

from mne.viz import use_browser_backend

# Load Raw
sample_data_folder = mne.datasets.sample.data_path()
raw_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                        'sample_audvis_raw.fif')
raw = mne.io.read_raw(raw_file)

# Add test-annotations
annotations = mne.Annotations([1, 10, 20, 30],
                              [1, 2, 1, 2],
                              ['a', 'b', 'a', 'b'])

with use_browser_backend('qt'):
    browser = raw.plot(events=events, block=True, title='Sample-Data',
                       clipping=None, show=True, precompute=True,
                       use_opengl=True, theme='auto')
```
Then remove all descriptions from the plot in annotation mode and hit "Remove description" again.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2375,27 +2375,28 @@ class AnnotationDock(QDockWidget):
                 icon=QMessageBox.Information)
 
     def _remove_description(self, rm_description):
-        # Remove regions
-        for rm_region in [r for r in self.mne.regions
-                          if r.description == rm_description]:
-            rm_region.remove()
+        if rm_description != "":
+            # Remove regions
+            for rm_region in [r for r in self.mne.regions
+                              if r.description == rm_description]:
+                rm_region.remove()
 
-        # Remove from descriptions
-        self.mne.new_annotation_labels.remove(rm_description)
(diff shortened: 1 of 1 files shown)


mscheltienne:
I can not reproduce neither of this bugs on `main`. Could you detail a bit more?
For the warning-box, it's behind this clause:

```
if self.mne.current_description:
    # handle drag event
elif event.isFinish():
    # warning box since we finished without a description being set
```

And `self.mne.current_description` is set to None when you don't have a description set, so that part should already be covered on main.

marsipu:
@mscheltienne I added how to reproduce the remove-description one.
The other I actually can't reproduce at the moment either. But I had it this morning, can't explain why.

mscheltienne:
The remove-description one only occurs when `block=True`, interesting. Anyway, the change to `_remove_description` makes sense to me. The other one, not as much.

PR #181, opened 2023-08-01
ENH: Improve vline

marsipu:
#### What does this implement/fix?

This changes the width of the vline to improve visibility as suggested in #155.
It also adds that the vline-label changes color too when the mouse hovers over the vline.

@larsoner @mscheltienne WDYT should be the width (examples in #155)?

PR #184, opened 2023-08-03
Add missing update of the label of a VLine

mscheltienne:
@marsipu Following #181, that's what was causing the weird hovering behavior on my side.


marsipu:
Great thanks @mscheltienne , I could reproduce it on Linux too now.

marsipu (approved):

PR #185, opened 2023-08-03
[DO NOT MRG] Run CIs on main and on mscheltienne/mne-python:annotations

mscheltienne:
To confirm that everything is working in https://github.com/mne-tools/mne-qt-browser/pull/178, running:

- `mne-qt-browser` main with `mne` main: Should work.
- `mne-qt-browser` main with https://github.com/mne-tools/mne-python/pull/11855: Should fail the added test.

### DIFF
--- a/.github/workflows/qt_viz_tests.yml
+++ b/.github/workflows/qt_viz_tests.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, windows, macos]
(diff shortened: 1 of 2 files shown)


mscheltienne:
Hmm unless, I'm mistaken, that's actually already run as part of https://github.com/mne-tools/mne-python/pull/11855 so that PR to change the CIs here and test mne-qt-browser main vs mme-python:annotations is useless. Oops.

PR #187, opened 2023-08-04
Apply the update to the projectors on bad channel toggle

### LINKED ISSUE #186: Update projectors when bad channels are selected/unselected
Probably not the simplest chain of events, but we had recently a bad mag channel that was spreading artifacts (DC jumps) to other mags through the SSP. It was a bit difficult to find which one was responsible because when you select a trace to set the channel as bad, the signal in the view area is not updated. To update it, I was pressing right/left arrow to change the x-axis selection. I guess it would have also work if I change the Y-axis channel selection.

There is probably a missing update call after setting a channel as "bad", update which is performed when the X-axis selection is changed.

mscheltienne:
Closes #186

The projectors were updated, but the update was not applied. Likely the pre-computation has to be re-run.
Anyway, that small change fixes the issue. It does however mean that we run twice through `self._update_projector()` because `_toggle_bad_channel` defined in `mne-python` runs it as well, but the cost is small.

Also, I noticed an import of the Qt5 backend at the top of the file. I'm guessing that whatever functionality uses it is currently failing with a Qt6 binding, but I did not test this. We had a similar import in `mne-icalabel` that failed when we started using 22.04 LTS + `qt6-base-dev` instead of 20.04 LTS + `qt5-default` (https://github.com/mne-tools/mne-icalabel/pull/120).

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2053,9 +2053,7 @@ class AnnotRegion(LinearRegionItem):
         # remove merged regions
         overlapping_regions = list()
         for region in self.mne.regions:
-            if region.description != self.description:
-                continue
-            if id(self) == id(region):
+            if region.description != self.description or id(self) == id(region):
                 continue
             values = region.getRegion()
             if any(self.getRegion()[0] <= val <= self.getRegion()[1] for val in values):
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -37,7 +37,7 @@ from qtpy.QtWidgets import (QAction, QColorDialog, QComboBox, QDialog,
                             QGraphicsLineItem, QGraphicsScene, QTextEdit,
                             QSizePolicy, QSpinBox, QSlider, QWidgetAction,
                             QRadioButton)
-from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg
+from matplotlib.backends.backend_qtagg import FigureCanvasQTAgg
 from matplotlib.colors import to_rgba_array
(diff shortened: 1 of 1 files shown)


drammock (approved):

drammock:
> The projectors were updated, but the update was not applied. Likely the pre-computation has to be re-run.

Does this mean that if you scroll, it's showing data with the *previous* projector (before the bad channel toggle)?  Does calling `_apply_update_projector` trigger the precompute to be redone?

> Anyway, that small change fixes the issue. It does however mean that we run twice through `self._update_projector()` because `_toggle_bad_channel` defined in `mne-python` runs it as well, but the cost is small.

I wonder if this means the Qt browser helper methods aren't optimal (i.e., maybe needs to split "apply projector" helper into separate "update" and "apply" steps?)

mscheltienne:
I didn't check in details, but I'm 90% sure the issue was the precomputation. `_apply_update_projector` does trigger a rerun of the precomputation, while this is nt the case of `self._update_projector`. 

> Qt browser helper methods aren't optimal

Definitely not!! I found a lot of different scenarios in which some functions/codes would be executed either for no reason or multiple times. Most of the time, it's due to signals emitted because of a `.setXXX` or equivalent method in the browser. I added a signal blocker in https://github.com/mne-tools/mne-qt-browser/pull/172 but it's barely used through the codebase for now. Hopefully, most of those scenarios don't yield an excessive overhead.

drammock:
> I didn't check in details, but I'm 90% sure the issue was the precomputation. `_apply_update_projector` does trigger a rerun of the precomputation, while this is nt the case of `self._update_projector`.

OK that makes sense. Merging; let's not try to fix everything all at once.

PR #195, opened 2023-09-12
in1d -> isin

drammock:
deal with numpy deprecation of `in1d`


drammock:
we really need to coordinate better to avoid duplicated effort.

larsoner:
Yes I should have commented as soon as I saw the problem in this repo, sorry! But at least it was hopefully only a few minutes of duplicated effort in this case

PR #199, opened 2023-09-21
Update pyproject.toml adding pyqt5 binding

timonmerk:
When installing mne-qt-browser under Windows 11, mne version 1.5.1, `pip install mne-qt-browser` does not install the required qt binding.

`raw.plot()` therefore results in the error:

> No Qt bindings could be found
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "C:\Users\ICN_admin\Anaconda3\envs\bispectra\lib\site-packages\mne\io\base.py", line 1825, in plot
>     return plot_raw(
>   File "<decorator-gen-202>", line 12, in plot_raw
>   File "C:\Users\ICN_admin\Anaconda3\envs\bispectra\lib\site-packages\mne\viz\raw.py", line 400, in plot_raw
>     fig = _get_browser(show=show, block=block, **params)
>   File "C:\Users\ICN_admin\Anaconda3\envs\bispectra\lib\site-packages\mne\viz\_figure.py", line 698, in _get_browser
>     fig = backend._init_browser(**kwargs)
> AttributeError: 'NoneType' object has no attribute '_init_browser'

Installing pyqt5 with `pip installl pyqt5` fixes the issue as described [here](https://github.com/spyder-ide/spyder/issues/3545#issuecomment-254049249). Therefore I added the dependency to the `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "pyopengl; platform_system=='Darwin'",
     "darkdetect",
     "qdarkstyle",
+    "pyqt5"
 ]
 dynamic = ["version"]
 


larsoner:
I don't think we want to do this. In principle mne-qt-browser can use PyQt5, PyQt5, PySide2, or PySide6, hence why we don't force users to install any specific one.

mscheltienne:
The error message raised when importing the browser should explicitly tell the user that he needs to choose and install one of the bindings, PyQt5, PyQt6, PySide2, or Pyside6. Is that not the case?

larsoner:
... and if it's not, I consider it a `qtpy` bug and we should open an issue there since we really rely on them to do all the binding business

hoechenberger:
> The error message raised when importing the browser should explicitly tell the user that he needs to choose and install one of the bindings, PyQt5, PyQt6, PySide2, or Pyside6.

I agree we should produce a helpful error message

mscheltienne:
How about a post-install script which checks for the presence of at least one binding, and else install a default one? That has to be doable with `setuptools`?

cbrnr:
TBH, I would try to improve the error message in QtPy. Installing something with a post-install script could lead to problems (I'm not an expert, but many package dependency analysis tools will probably not be able to parse such custom post-install things).

Another alternative would be to ditch QtPy and use a specific binding. It looks like PyQt6 would be the best option, but I'm sure this will trigger other issues (most likely related to PyVista).

cbrnr:
A third option could be that our backend issues an appropriate error message (i.e. we catch the generic one from QtPy and turn it into something more useful).

larsoner:
Agreed QtPy would be the right place but who knows if they'll accept a fix based on https://github.com/spyder-ide/qtpy/pull/221 going stale. I'll close this PR and @timonmerk if you're up for it feel free to open another PR to improve our first `qtpy` import with a try/except. The annoying part is that they raise PythonQtError, which is in their namespace... which you can only get if you can import `qtpy`. So you'll probably need something like:
```
try:
    from qtpy...
except Exception as exc:
    if exc.__class__.__name__ == "PythonQtError":
         raise ImportError(...) from None
```

PR #200, opened 2023-09-25
first pass at ch-specific annot

drammock:
pair-programmed with @scott-huberty

WIP, pending aesthetics discussion

PR #202, opened 2023-09-26
Vizualization framework for channel specific annotations

scott-huberty:
Closes #200
Closes #196 

(EDIT) starts to close https://github.com/mne-tools/mne-python/issues/8946 . I think we will need to implement the same funcitionality for the matplotlib backend before closing that issue.


scott-huberty:
This _almost_ works:

Of course, the sample audvis dataset always seems to present the most complex scenario (floating point sfreq, etc) 🙂

If I run this:


```
import mne
data_path = sample.data_path()
meg_path = data_path / "MEG" / "sample"
raw_fname = meg_path / "sample_audvis_raw.fif"
raw = mne.io.read_raw_fif(raw_fname, preload=True)
raw.pick("meg").filter(1, 30)

ch_names = ["MEG 0133", "MEG 0142", "MEG 0143", "MEG 0121"]
onsets = [1, 4.25, 7.25]
durs = [2, 1.75, 1.25]
descs = ["Specific_channels_test", 'BAD_test', 'Good_test']
chs = [ch_names, [], []]
annots = mne.Annotations(onset=onsets, duration=durs, description=descs, ch_names=chs)

raw.set_annotations(annots)
raw.plot(theme="light")

```

I get this plot

Screen Shot 2023-09-26 at 3 19 47 PM


"MEG 0143"  is in `ch_names`, but it is NOT highlighted in the plot. ["MEG 0212", "MEG 0222"] are NOT in the `ch_names` but they ARE highlighted. @drammock I think this has to do with our "dirty hack", and that these MEG channels are not ordered in the plot by their number (i.e. "MEG 0133", "MEG 0143", "MEG 0142").

drammock:
ah, right you are. Our dirty hack will not work. Maybe check if the channel order is separately specified / kept track of somewhere in `fig.mne`)?

larsoner:
Aesthetically pleasing despite the indexing bug!

For the indexing I would look at how we figure out when you click on a channel name Y label which channel is actually bad. Indexing correspondence between channel names and what is currently shown should be used there I think

larsoner:
@scott-huberty did you test interactivity? Using this code:
```
import mne
raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() / "MEG" / "sample" / "sample_audvis_raw.fif")
raw.annotations.append(raw.first_time + 1, 2, "CH_SPEC", [["MEG 0113"]])
raw.plot()
```
pressing "a" on the keyboard to enter annot mode and shrinking the annot I see the ch-specific one does not shrink:

Screenshot 2023-09-29 at 1 52 56 PM

Can you replicate?

scott-huberty:
I did not test that, and can replicate 😅

Looking into it

larsoner:
Code looks good other than the interactivity issue!

larsoner:
FYI @scott-huberty I changed your wording in the top-comment from "supersedes" to "closes" because the latter does GitHub closing magic

scott-huberty:
Going to stop for the day soon, but FYI the fix that we thought we had for this (updating the `xData` of the fillBetween items)  is not actually working (the fillBetween item does not redraw to update its location on the plot). to be continued...

larsoner:
xData was just a list, right? There is probably a .update or something that you need to call. But simpler would probably be too call https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/plotcurveitem.html#pyqtgraph.PlotCurveItem.setData on the two curves that define the fill between. If that doesn't fix it, calling that then https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/fillbetweenitem.html#pyqtgraph.FillBetweenItem.setCurves should do it. It's likely that these `set*` functions update some attributes under the hood that are not updated when you just assign a new value in the xData list.

scott-huberty:
Exactly! The `setData` is what we'll need.I just need to put it in the right place in the codebase. (I tried putting it in the `_stop_changed` method, but `stop_changed` isn't being triggered when I adjust an annotation)

scott-huberty:
the `FillBetweenItems` (i.e. shaded rectangles) of channels specific annotations will now stay in sync with the annotation region IF you use the start/stop spinboxes to change the annotation region start/stop time. For some reason, dragging or extending an annotation directly in the plotting area triggers a different code path. 


I had to rebase with main after #203, and force push my changes to this remote branch. Before doing that I made a local copy of the state of this remote branch, in case I realize that i've done something terribly wrong..

larsoner:
> in case I realize that i've done something terribly wrong..

At the end of the day almost all that matters for our squash-and-merge is the `diff` on GitHub, so just check to make sure that looks okay and you should be good

larsoner:
> For some reason, dragging or extending an annotation directly in the plotting area triggers a different code path.

This I think is where the `logger.debug` can/should help us. I'd add `logger.debug`s in all the click event handlers (now if you want!) and it should become clear which one is triggered by dragging the edge of an annot

scott-huberty:
Alright that should do it... I'll leave this in your court @larsoner @drammock - enjoy the dev retreat!

mne_qt_browser/_pg_figure.py
```diff
@@ -2206,24 +2225,60 @@ def _region_changed(self):
...
         # remove overlapping regions
         for region in overlapping_regions:
             self.weakmain()._remove_region(region, from_annot=False)
         # re-set while blocking the signal to avoid re-running this function
         with SignalBlocker(self):
             self.setRegion((onset, offset))
         self.update_label_pos()
+        # Update the FillBetweenItem shapes for channel specific annotations
+        if self.ch_annot_fills:
+            self._update_channel_annot_fills(onset, offset)
```
larsoner, line 2238: Let the iterator just return on its own (be a no-op) rather than checking its length
```suggestion
        self._update_channel_annot_fills(onset, offset)
```

mne_qt_browser/_pg_figure.py
```diff
@@ -2206,24 +2225,60 @@ def _region_changed(self):
...
-    def update_color(self):
-        """Update color of annotation-region."""
+    def _update_channel_annot_fills(self, start, stop):
+        """Update the FillBetweenItems for channel specific annotations.
+
+        FillBetweenItems are used to highlight channels associated with an annotation.
+        Start and stop are time in seconds.
+        """
+        logger.debug(f"moving {self.description} rectangles to {start} - {stop}")
+        for this_fill in self.ch_annot_fills:
```
larsoner, line 2247:
```suggestion
        for fi, this_fill in enumerate(self.ch_annot_fills):
        if fi == 0:
            logger.debug(f"moving {len(self.ch_annot_fills} {self.description} rectangle(s) to {start} - {stop}")
```

mne_qt_browser/_pg_figure.py
```diff
@@ -2690,6 +2754,9 @@ def _stop_changed(self):
         start = sel_region.getRegion()[0]
         if start < stop:
             sel_region.setRegion((start, stop))
+            # Make channel specific fillBetweens stay in sync with annot region
+            if sel_region.ch_annot_fills:
+                sel_region._update_channel_annot_fills(start, stop)
```
larsoner, line 2759:
```suggestion
            sel_region._update_channel_annot_fills(start, stop)
```

mne_qt_browser/tests/test_pg_specific.py
```diff
@@ -85,6 +85,58 @@ def test_annotations_interactions(raw_orig, pg_backend):
     fig.msg_box.close()
 
 
+def test_ch_specific_annot(raw_orig, pg_backend):
+    """Test plotting channel specific annotations."""
+    from mne import Annotations
+    from pyqtgraph.graphicsItems.FillBetweenItem import FillBetweenItem
+
```
larsoner, line 92: Shouldn't nest in tests

larsoner (approved):
I'll commit these and merge, thanks @scott-huberty !

PR #205, opened 2023-10-02
Improve error message if qt binding could not be found

timonmerk:
Related to #199, this PR checks before the first qtpy import if the `QtBindingsNotFoundError` exception is thrown. The error message now hints that pyqt5 needs to be installed.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -23,18 +23,7 @@ from os.path import getsize
 import numpy as np
 
 try:
-    from qtpy.QtCore import (
-        QEvent,
-        QThread,
-        Qt,
-        Signal,
-        QRectF,
-        QLineF,
-        QPointF,
-        QPoint,
-        QSettings,
-        QSignalBlocker,
-    )
+    from qtpy.QtCore import Qt
 except Exception as exc:
     if exc.__class__.__name__ == "QtBindingsNotFoundError":
         raise ImportError(
@@ -43,6 +32,18 @@ except Exception as exc:
     else:
         raise
 
+ from qtpy.QtCore import (
+    QEvent,
+    QThread,
+    Qt,
+    Signal,
+    QRectF,
+    QLineF,
+    QPointF,
+    QPoint,
+    QSettings,
+    QSignalBlocker,
+)
 from qtpy.QtGui import (
     QFont,
     QIcon,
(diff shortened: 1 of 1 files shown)


mne_qt_browser/_pg_figure.py
```diff
@@ -21,18 +21,26 @@
...
+        QRectF,
+        QLineF,
+        QPointF,
+        QPoint,
+        QSettings,
+        QSignalBlocker,
+    )
+except Exception as exc:
+    if exc.__class__.__name__ == "QtBindingsNotFoundError":
+        raise ImportError("No Qt binding found, please install pyqt5") from None
```
larsoner, line 40:
```suggestion
        raise ImportError("No Qt binding found, please install pyqt5") from None
```
```suggestion
        raise ImportError("No Qt binding found, please install PyQt6, PyQt5, PySide6, or PySide2") from None
```

mne_qt_browser/_pg_figure.py
```diff
@@ -21,18 +21,26 @@
...
+    from qtpy.QtCore import (
+        QEvent,
+        QThread,
+        Qt,
+        Signal,
+        QRectF,
+        QLineF,
+        QPointF,
+        QPoint,
+        QSettings,
```
larsoner, line 35: I would also just nest one of these imports, then the rest can be as they were before. It'll make the `diff` smaller as well

mne_qt_browser/_pg_figure.py
```diff
@@ -21,18 +21,26 @@
...
+        QSignalBlocker,
+    )
+except Exception as exc:
+    if exc.__class__.__name__ == "QtBindingsNotFoundError":
+        raise ImportError("No Qt binding found, please install pyqt5") from None
+    else:
+        raise exc
```
larsoner, line 42: IIRC the traceback is better this way, no extra stuff from reraising
```suggestion
        raise
```

mne_qt_browser/_pg_figure.py
timonmerk: Thanks for the comments @larsoner

mne_qt_browser/_pg_figure.py
```diff
@@ -21,10 +21,20 @@
...
+    from qtpy.QtCore import Qt
+except Exception as exc:
+    if exc.__class__.__name__ == "QtBindingsNotFoundError":
+        raise ImportError(
+            "No Qt binding found, please install PyQt6, PyQt5, PySide6, or PySide2"
+        ) from None
+    else:
+        raise
+
+ from qtpy.QtCore import (
```
larsoner, line 35:
```suggestion
from qtpy.QtCore import (
```

larsoner (approved):
Marking for merge when green, thanks in advance @timonmerk !

PR #208, opened 2023-10-20
Sensitivity Scaling - Deprecated

Grifunf:
#### Reference issue

Closes mne-tools/mne-python#10888

This supports additional functionalities for the Sensitivity (amplitude) scalings of data traces:
- Scale values from drop-down menu
- Scale Values from manual user input
- Scale Values for specific channel types of the viewed file


drammock:
sorry for the slow response on this one! I played around with this a little, and here's what I noticed:

1. the "advanced amplitude" settings dialog has a second tab called "channels" which is empty.
2. the values shown in the advanced amplitude dropdown boxes don't necessarily reflect what is being shown on screen: it is possible to click the zoom in / zoom out toolbar buttons while the advanced dialog is open, and if you do that the signal traces change but the dropdown boxes do not change. This seems misleading to me.
3. Given that the goal is to set amplitude scales relative to one centimeter of monitor height, I think the magenta scale bars on the plot should probably be made 1cm tall. Based on the values in the amplitude settings dropdowns and the values shown on the scalebars, this isn't the case (or at least not consistently so)

More importantly, if I look back at [the recommended sequence of steps](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1712138949) from the issue discussion, it seems like you're not working through them in order:

> 1. add UI elements that display the current scaling for each channel type (one per channel type in the viewed file)
> 2. in the UI elements from (1), let users type in a new scaling to change it for a given channel type.
> 3. Add more UI elements to display the current sensitivity for each channel type. These would be read-only for now, but at least users could use the +/- keys to get close to the sensitivity that they want
> 4. in the UI elements from (3), let users type in a new sensitivity to change it for a given channel type.
> 5. Design a nicer UI for this feature rather than 2 columns (or rows) of text boxes. Maybe this means dropdowns or comboboxes, but for now let's focus on getting it to work, then at the end we can worry about making it pretty / easy to use.

It seems like you've skipped step 1, 2, 3, and *sort of* done step 4 (users choose pre-defined values instead of typing in arbitrary values). But it also seems like you're not yet actually detecting DPI / calculating actual screen size? I don't see `dpi` anywhere in the PR and [your comment today on the issue thread](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1789053348) seems to confirm that.

I will reiterate [what I said before](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1762241854):

> The best advice I can give is to work on [the steps I outlined previously](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1712138949), one at a time, in order

This will make it *really really really* quite a lot easier to troubleshoot and review your work.

Grifunf:
@drammock  First of all, I'm really sorry, I have indeed wandered off your given steps and there are certain things that need clearing up:

> the "advanced amplitude" settings dialog has a second tab called "channels" which is empty.

The first idea was to have seperate scaling for All, Types & Channels, but it was abandoned. This will be removed.
> the values shown in the advanced amplitude dropdown boxes don't necessarily reflect what is being shown on screen

This is something I wanted to discuss. For the time being, I didn't replace the existing scaling factor, but rather multiplied it with the new one, so the 2 may coexist. This does bring up some problems, one being the one you pointed out, the other being the opposite. If you first press the zoom in/out buttons and then edit the dropboxes, it scales but not to the desired value.
> It seems like you've skipped step 1, 2, 3, and sort of done step 4 (users choose pre-defined values instead of typing in arbitrary values). But it also seems like you're not yet actually detecting DPI / calculating actual screen size?

I indeed haven't followed your steps exaclty, and would like to clear some things up:

1. If I understand correctly, steps 1&2 about scaling are about _time_ scaling, whereas steps 3&4 about sensitivity are about _amplitude_ scaling. I had already started with the amplitude, so I would just do those two in reverse order.
2. In my starting plan, the first step would be to design the UI/Editing system, and the final step would be to integrate the monitor size. The current labels are indeed misleading (sorry about that) and were written with the final step in mind.
3. I skipped the UI elements part and went straight to the editability, because the Scalebars already display the current sensitivity of each channel type. So in my understanding step 3 wasn't necessary. That isn't the case for the time scaling though (step 1) where I will implement it.
4. Also what do you mean instead of typing arbitrary values? They are editable, apart from the pre-defined values

I will come back soon with the progress you asked for. For the time being, I would like to hear your thoughts on the above, as well as ask: Do we really need textboxes on the toolbar? Isn't a small dialog with textboxes better? Also, I would like to persist on my [last 2 questions](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1789053348), as I feel I am going to need them sooner or later.
Thanks again and sorry for messing things up, it's my first major project as you can tell :)

drammock:
> 1. If I understand correctly, steps 1&2 about scaling are about _time_ scaling, whereas steps 3&4 about sensitivity are about _amplitude_ scaling.

No, all steps are about amplitude scaling. For example, step 1 talks about doing things "for each channel type" but we never intended to scale *time* differently for each channel type.

> In my starting plan, the first step would be to design the UI/Editing system

See my earlier comments about not doing the part that seems easiest first. Designing the UX *before* we've worked out the kinks in actually computing sensitivity and converting between sensitivity and display coordinates (1) is risky for the reasons mentioned before, and (2) ignores the fact that we've sort of "punted" on discussing the UX among the maintainer team, on the assumption that we could decide on the best UX later after we knew that the new feature worked. In other words, you're implementing *your vision* of what the UX should be like, without having established a consensus among other stakeholders on what it should look like.

> I skipped the UI elements part and went straight to the editability, because the Scalebars already display the current sensitivity

No, not really; the scalebars display the *scale*.  "Sensitivity" is defined as e.g. microvolts *per centimeter* isn't it?  And the user has no information about how tall the scalebar is.

>  what do you mean instead of typing arbitrary values? They are editable, apart from the pre-defined values

That may be my mistake, I probably didn't try to edit them / assumed because there were pre-populated values that it was a selection box and not a combo box.

Please don't mistake my tone if it comes across as terse or angry. I am happy that you're working on this!  And I want to see it get merged in.  But for that to happen, multiple people need to review the work, and they're all busy, so I will continue to be insistent that you follow the recommended steps *in order* and check in with us in between steps, so that we can (1) review *small* diffs of related changes that are easy to understand and make it easy to spot bugs or improvements, and (2) prevent you from going down blind alleys or otherwise needing to un-do or re-do things.

drammock:
>  Do we really need textboxes on the toolbar? Isn't a small dialog with textboxes better?

This is meant to be a "quick and dirty" approach that will get revised later. See item (5): "Design a nicer UI for this feature rather than 2 columns (or rows) of text boxes."

> I tried some for-loop and enumerate iterations but it has become incredibly slow. Do you have any advice on how to tackle this?

If you start with step 1, you don't need to deal with this yet. I will not help debug or optimize that particular part of your code until we've reached the point where it's necessary.

> I was wondering whether we already have access to the monitor DPI or something relevant.

- https://stackoverflow.com/questions/3129322/how-do-i-get-monitor-resolution-in-python
- monitor size in pyqt5: https://stackoverflow.com/questions/35887237/current-screen-size-in-python3-with-pyqt5
- DPI in pyqt5: https://stackoverflow.com/a/54273855/1664024

Grifunf:
Thanks a lot! One final question, because I don't seem to get it:
What's the difference between scaling and sensitivity? If you could provide an example I would be grateful.

drammock:
> Thanks a lot! One final question, because I don't seem to get it: What's the difference between scaling and sensitivity? If you could provide an example I would be grateful.

`scalings` is a parameter passed to many of our plotting functions, and broadly speaking it determines the y-axis scale. If you passed `scalings=1` then the data would be plotted exactly as it's represented in memory, which means SI units (Volts for EEG, Teslas for magnetometers, etc). If you did this, the y-axis values would be really really small since the typical dynamic range for EEG is on the order of *micro*Volts and for magnetometers it's *femto*Teslas.  So `scalings` basically just multiplies the signals before plotting them to make the y-axis tick labels look like `2, 3, 4` instead of `0.000002, 0.000003, 0.000004`, and changes the unit in the y-axis label.

When it comes to `raw.plot()` it's a bit more complicated because we don't draw y-axis ticks (or rather, the y-axis tick labels are the channel names, not scale values).  But we still use `scalings` --- we change how big the signal is drawn instead of changing the numbers on the tick labels. Each signal is drawn in a rectangle that is 1 unit tall in plot coordinates, and by default that reflects 1 Volt for EEG, 1 Tesla for magnetometers, etc. --- except that by default, `scalings` does not equal 1, it equals [values chosen to usually yield nice, readable plots for typical signals of each sensor type](https://github.com/mne-tools/mne-python/blob/e4a6eba5d0a4c5f92b904207c33252f1979043db/mne/defaults.py#L129-L130).

None of what I said above takes into account screen size, screen DPI, or even plot window size. Right now, if you reduce the height of your plot window by half, the signal traces will shrink by about half too. "Sensitivity" is a way of specifying the actual physical size of the signal trace on screen, which is why it is in terms of "Volts *per meter*" or "microVolts *per centimeter*".  If our GUI handled sensitivity correctly, then resizing the figure window would *not* cause the signal traces to grow or shrink in height.

it may help to read the original thread, starting from the big screenshot: https://mne.discourse.group/t/scaling-sensitivity-uv-mm/5079/12

hoechenberger:
I'm wondering if we could come up with a more expressive name for this feature, other than Sensitivity?

drammock:
> I'm wondering if we could come up with a more expressive name for this feature, other than Sensitivity?

apparently that's what it's called in the clinical setting, which is where the feature request comes from. I think changing the name makes it less discoverable for the people who would want it.

hoechenberger:
Good point, let's stick with sensitivity then.

larsoner:
@Grifunf feel free to try latest `main` of this repo, it should hopefully have equivalent functionality for you now!

PR #212, opened 2023-11-04
Scaling Text Boxes

Grifunf:
#### Reference issue

Closes mne-tools/mne-python#10888

Second toolbar with text boxes that display the `scalings` values for the channel types of the file viewed. Currently not responding to user interaction.

Test that asserts that the default and some custom `scalings` values from the `plot_raw()` call are passed to the textboxes correctly.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3518,7 +3518,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             lbl = QLabel(chtype.upper())
             self.mne.toolbar2.addWidget(lbl)
             box = QLineEdit()
-            box.setPlaceholderText(str(self.mne.scalings[chtype]))
+            box.setText(str(self.mne.scalings[chtype]))
             self.scale_boxes[chtype] = box
             self.mne.toolbar2.addWidget(box)
 
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3509,6 +3509,19 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         ahelp.triggered.connect(self._toggle_help_fig)
         self.mne.toolbar.addAction(ahelp)
 
+        # 2nd Toolbar accommodating the Text Boxes of Scalings
+        self.addToolBarBreak()
+        self.mne.toolbar2 = self.addToolBar("Scales")
+        # Scalings Text Boxes
+        self.scale_boxes = OrderedDict()
+        for chtype in [ct for ct in self.mne.ch_types_ordered if ct != "stim"]:
+            lbl = QLabel(chtype.upper())
+            self.mne.toolbar2.addWidget(lbl)
+            box = QLineEdit()
+            box.setPlaceholderText(str(self.mne.scalings[chtype]))
+            self.scale_boxes[chtype] = box
+            self.mne.toolbar2.addWidget(box)
+
         # Set Start-Range (after all necessary elements are initialized)
         self.mne.plt.setXRange(
             self.mne.t_start, self.mne.t_start + self.mne.duration, padding=0
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3133,6 +3133,12 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         self.mne.traces = list()
         # Scale-Factor
         self.mne.scale_factor = 1
+        # Ordered channel types list, used in multiple instances
+        ordered_types = self.mne.ch_types[self.mne.ch_order]
+        unique_type_idxs = np.unique(ordered_types,
+                                     return_index=True)[1]
+        self.mne.ch_types_ordered = [ordered_types[idx] for idx
+                            in sorted(unique_type_idxs)]
         # Stores channel-types for butterfly-mode
         self.mne.butterfly_type_order = [
             tp for tp in DATA_CH_TYPES_ORDER if tp in self.mne.ch_types
@@ -3731,12 +3737,9 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         """
         self.mne.scalebars.clear()
         # To keep order (np.unique sorts)
-        ordered_types = self.mne.ch_types[self.mne.ch_order]
-        unique_type_idxs = np.unique(ordered_types, return_index=True)[1]
-        ch_types_ordered = [ordered_types[idx] for idx in sorted(unique_type_idxs)]
         for ch_type in [
             ct
-            for ct in ch_types_ordered
+            for ct in self.mne.ch_types_ordered
             if ct != "stim"
             and ct in self.mne.scalings
             and ct in getattr(self.mne, "units", {})
--- a/mne_qt_browser/tests/test_pg_specific.py
+++ b/mne_qt_browser/tests/test_pg_specific.py
@@ -19,7 +19,25 @@ INCREASE_AMPLITUDE = "Increase amplitude"
 TOGGLE_ANNOTATIONS = "Toggle annotations mode"
 SHOW_PROJECTORS = "Show projectors"
 
+def test_scalings(raw_orig, pg_backend):
+    """Test the correct parsing of the scalings dict to Text Boxes"""
+    fig = raw_orig.plot(scalings = 'auto')
+    fig.test_mode = True
+    QTest.qWaitForWindowExposed(fig)
+    QTest.qWait(50)
+    for type in [ct for ct in fig.mne.ch_types_ordered if ct != "stim"]:
+        assert fig.scale_boxes[type].text() == str(fig.mne.scalings[type])
+    fig.close()
+    fig = raw_orig.plot()
+    fig.test_mode = True
+    QTest.qWaitForWindowExposed(fig)
+    QTest.qWait(50)
+    for type in [ct for ct in fig.mne.ch_types_ordered if ct != "stim"]:
+        assert fig.scale_boxes[type].text() == str(fig.mne.scalings[type])
+    fig.close()
 
+
+'''
 def test_annotations_interactions(raw_orig, pg_backend):
     """Test interactions specific to pyqtgraph-backend."""
     # Add test-annotations
@@ -433,3 +451,4 @@ def test_pg_toolbar_actions(raw_orig, pg_backend):
     assert pg_backend._get_n_figs() == 2
     fig._fake_click_on_toolbar_action("Help", wait_after=100)
     assert pg_backend._get_n_figs() == 1
+'''
--- a/mne_qt_browser/tests/test_pg_specific.py
+++ b/mne_qt_browser/tests/test_pg_specific.py
@@ -21,19 +21,21 @@ SHOW_PROJECTORS = "Show projectors"
 
 def test_scalings(raw_orig, pg_backend):
     """Test the correct parsing of the scalings dict to Text Boxes"""
-    fig = raw_orig.plot(scalings = 'auto')
-    fig.test_mode = True
-    QTest.qWaitForWindowExposed(fig)
-    QTest.qWait(50)
-    for type in [ct for ct in fig.mne.ch_types_ordered if ct != "stim"]:
-        assert fig.scale_boxes[type].text() == str(fig.mne.scalings[type])
-    fig.close()
     fig = raw_orig.plot()
     fig.test_mode = True
     QTest.qWaitForWindowExposed(fig)
     QTest.qWait(50)
-    for type in [ct for ct in fig.mne.ch_types_ordered if ct != "stim"]:
+    for type in fig.scale_boxes.keys():
         assert fig.scale_boxes[type].text() == str(fig.mne.scalings[type])
+    fig.close()
+    scalings = dict(mag=1.0, grad=2.0, eeg=3.0, eog=4.0)
+    fig = raw_orig.plot(scalings = scalings)
+    fig.test_mode = True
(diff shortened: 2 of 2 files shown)


hoechenberger:
Could you share a screenshot please? 🙏

Grifunf:
Ok, so I think this is pretty much step 1 completed. I wait to hear any comments, but I guess I am moving on to step 2, the user interaction and related tests with the Textboxes.

Grifunf:
> Could you share a screenshot please? 🙏

Ofc, sorry
image

Grifunf:
I would also like to mention that there are weird failures on the speed tests, even on the upstream version.
```
    fig = epochs.plot(show=False, block=False, events=False, **benchmark_param)
d:\repos\mne-python\mne\epochs.py:1306: in plot
    return plot_epochs(
d:\repos\mne-python\mne\viz\epochs.py:950: in plot_epochs
    event_nums = events[:, 2]
E   TypeError: 'bool' object is not subscriptable</failure></testcase><testcase classname="mne_qt_browser.tests.test_speed" name="test_scroll_speed_epochs_unicolor[use_opengl=True]" time="0.458"><failure message="TypeError: 'bool' object is not subscriptable">test_speed.py:155: in test_scroll_speed_epochs_unicolor
```
Should I be concerned by this?

larsoner:
+1 for having display + interaction in this PR. I think it's a natural thing to add. FWIW I'd expect input validation on the text boxes like for `GRAD: 4e-11` if I type `"foo"` and hit enter it should go back to `4e-11`, but something like `34.1` should work. I'm pretty sure there is a way to do this with Qt. 

For the labels, though, instead of using `ch_type.upper()` or whatever it is now, you should use `DEFAULTS["titles"].get(ch_type, ch_type.upper())` or similar to get nice human readable ones like we do here

https://github.com/mne-tools/mne-python/blob/55b0184b94ec1c3d6b2559eeed383e760220f516/mne/_fiff/meas_info.py#L1651
https://github.com/mne-tools/mne-python/blob/55b0184b94ec1c3d6b2559eeed383e760220f516/mne/_fiff/meas_info.py#L1710

mne_qt_browser/tests/test_pg_specific.py
```diff
@@ -20,6 +20,26 @@
 SHOW_PROJECTORS = "Show projectors"
 
 
+def test_scalings(raw_orig, pg_backend):
+    """Test the correct parsing of the scalings dict to Text Boxes"""
```
drammock, line 24: Our docstring formatting checker is very picky (it requires full sentences with period)

```suggestion
    """Test the correct parsing of the scalings dict to Text Boxes."""
```

mne_qt_browser/_pg_figure.py
```diff
@@ -3503,6 +3509,19 @@ def __init__(self, **kwargs):
         ahelp.triggered.connect(self._toggle_help_fig)
         self.mne.toolbar.addAction(ahelp)
 
+        # 2nd Toolbar accommodating the Text Boxes of Scalings
+        self.addToolBarBreak()
+        self.mne.toolbar2 = self.addToolBar("Scales")
+        # Scalings Text Boxes
+        self.scale_boxes = OrderedDict()
+        for chtype in [ct for ct in self.mne.ch_types_ordered if ct != "stim"]:
+            lbl = QLabel(chtype.upper())
```
drammock, line 3518: Copying @larsoner's suggestion here: 

> For the labels, though, instead of using ch_type.upper() or whatever it is now, you should use DEFAULTS["titles"].get(ch_type, ch_type.upper()) or similar to get nice human readable ones like we do here

drammock (commented):
nice, very clean. Please proceed to add interactivity to the text boxes.

> I would also like to mention that there are weird failures on the speed tests, even on the upstream version.

Those tests are passing on the CIs (and for me locally). Probably they're failing locally for you because your local version of MNE-Python is outdated / doesn't yet have the bugfix for that particular failure

hoechenberger:
Hello, can we limit the width of those text boxes? They are waaaay to wide on my screen, which doesn't look good.
Screenshot 2023-11-08 at 18 01 24

drammock:
> Hello, can we limit the width of those text boxes? They are waaaay to wide on my screen, which doesn't look good.

This is [step 1 of a 5-step process](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1712138949).  Step 5 is "figure out a better UX". Please be patient.

Grifunf:
Ok, so I used a RegEx validator, because the doubles are a mess with the different locales (1,23 != 1.23).
Right now the floats are passed correctly to the `scalings` dict. I am currently writing the relevant test.

Grifunf:
So, in order to continue, I 'd like to yet again clear things up.

- First of all, right now nothing happens to the plot when a user changes the `scalings`. Nothing was supposed to happen anyways, because as we have mentioned before, the dict is currently irrelevant to the plot, apart from the [`plot_raw()`](https://mne.tools/stable/generated/mne.viz.plot_raw.html) argument. That's why in my first attempt I used a different approach. So now there is the question on how to continue.
- Furthermore, I want to discuss the next step: **sensitivity**. I think I still haven't quite got it. Please correct me if I'm wrong: In the sample file with the [above screenshots](https://github.com/mne-tools/mne-qt-browser/pull/212#issuecomment-1802299639), sensitivity would be the `800ft/cm` (or maybe `(800ft/cm)/cm` when taking into account the monitor size) scalebar value and is our end goal. So the concept is the user will be able to edit it, for example into `400ft/cm`, and the plot size will double in size (if you half its sensitivity). Sensitivity takes in account multiple things, including the `scalings`, and the `scale_factor`. In my first attempt, i created another arbitrary value to go into those calculations, enabling the control of the final result, the sensitivity.
- So for [step 3-4](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1712138949), I need to create another set of textboxes showing the current sensitivity value, which is gonna be the same as the scalebar values, and in continuation will be edited by the user.

So, I would like to know 2 things:

1. Have I gotten anything about sensitivity? Otherwise please correct me
2. What is the plan on how the user edits will affect the plot, either for `scalings` or for `sensitivity`? Right now I don't see any.

hoechenberger:
Nice work so far!

One question, it seems to me you're using text widgets, right? But since we're dealing with numbers, maybe a number widget would be more appropriate? Not sure if and how these support scientific notation though. But this could resolve your issue regarding the decimal symbol in a very natural way. And we could even add spin boxes

Grifunf:
> Nice work so far!
> 
> One question, it seems to me you're using text widgets, right? But since we're dealing with numbers, maybe a number widget would be more appropriate? Not sure if and how these support scientific notation though. But this could resolve your issue regarding the decimal symbol in a very natural way. And we could even add spin boxes

With some googling I found out that LineEdits with validators are the way to go. Customized my RegEx validator and now it works just fine on that part. No commas allowed though.

drammock:
> sensitivity would be the `800ft/cm` (or maybe `(800ft/cm)/cm` when taking into account the monitor size) scalebar value and is our end goal.

800 fT/cm would be a plausible *scaling* for gradiometers or a plausible *sensitivity* for magnetometers.  For gradiometers, sensitivity would be like 800 fT/cm/cm.

> right now nothing happens to the plot when a user changes the scalings. Nothing was supposed to happen anyways, because as we have mentioned before, the dict is currently irrelevant to the plot

I think though that this is the next thing to change behind the scenes. Make it so the vertical size of the traces is determined by *the value in the textbox* instead of *the value passed in the dict when the plot function was called* (of course the textboxes are populated from the dict, so *initially* the dict will still determine how things look).  This would also probably be the point where you start accounting for the plus/minus zoom buttons; i.e., making the scaling shown in the text box update when the zoom buttons are pressed.

After that, I think it's time to start reading monitor resolution/screen size/window size, and then after that, add an additional row of boxes that display *sensitivity* (not scaling), recompute sensitivity each time the scalings are changed, and update those text boxes with the new sensitivity.

mne_qt_browser/_pg_figure.py
```diff
@@ -42,6 +42,7 @@
     QPoint,
     QSettings,
     QSignalBlocker,
+    QRegExp,
```
mscheltienne, line 45: Instead of `QRegExp` and `QRegExpValidator`, if you want a regular expression based validator, please use `QRegularExpression` and `QRegularExpressionValidator` else it will be incompatible with Qt6.

Grifunf: Done

Grifunf:
@drammock 
>800 fT/cm would be a plausible scaling for gradiometers or a plausible sensitivity for magnetometers. For gradiometers, sensitivity would be like 800 fT/cm/cm.

Ok, this time I got it for sure.

Now moving on, I'd like to explain the way I'm going to implement this, in a little bit of detail, in order to make sure I'm in the right track and hear any inputs. This is btw the way I did it in my last botched up PR, where it ended up being functional, apart from a problem I had and will probably encounter again.

So, first of all, in my understanding, the `scalings` dict is used in the computing of the data outside of this module, upon the call of `plot_raw()`. Inside the `_pg_figure.py` module, the dict is used only for computing labels for the data (for example Scalebars), never actually modifying the data.

In order to perform scaling, as [this comment](https://github.com/Grifunf/mne-qt-browser-dev/blob/8325ea712b0e6f76a09dcc84e404d755b0e4f8ae/mne_qt_browser/_pg_figure.py#L3806C9-L3806C59) suggests: `# Scale Traces (by scaling the Item, not the data)`, the `scale_factor` value comes into play and it only affects the size of the traces, not the data itself. This makes sense in computational efficiency, especially since not every channel is plotted at once and a relatively small number of traces are used.

It seems to me that in order to achieve any sort of scaling (be it scaling or sensitivity), the creation of a `scale_factor_dict` is inevitable (one that holds different factors for different types). That was my first approach, where i multiplied every "global" `scale_factor` with the new `scale_factor_dict`. The results where the desired effects, but it didn't work correctly with the Zoom In/Out buttons. On my second approach, I think I will completely replace the `scale_factor` with the `scale_factor_dict`, and it will probably solve these problems smoothly.

That's all, sorry for the long comment. I would like to hear any suggestions or objections that you have on this. Meanwhile, I will try to do it.

Grifunf:
First of all sorry, the last commits have been a bit messy, I am working on it. Here's what i implemented so far:

- A dict named [`scale_factors`](https://github.com/Grifunf/mne-qt-browser-dev/blob/aa0c9f47fc2985253cf7813a433376d71fb0e3f8/mne_qt_browser/_pg_figure.py#L3147), holding the current scaling factor for each type. The values equal the `current_scalings / previous_scalings` calculation, and the results are ready for scaling, without the need to compute it constantly. (Note that the `scalings` dict isn't used in any calculation in the backstage)
- Similarly, a dict named [`norms_dict`](https://github.com/Grifunf/mne-qt-browser-dev/blob/aa0c9f47fc2985253cf7813a433376d71fb0e3f8/mne_qt_browser/_pg_figure.py#L3149), holding the calculations `scalings * unit_scalings` for each channel type. The calculation only needs to happen once in the startup, and not every time those blocks were called. So the dict initializes with the start values, and is further used with the `scale_factors`, as happened in the original version. 

I have one question. I see almost all tests fail, raising a `KeyError 'ias'`. This is a channel type that I presume is in a test file, that doesn't have a corresponding entry in the `scalings` dict (or the `unit_scalings`). I had the same problem with the `'stim'` type once, that is already excluded in multiple types iterations in the original version of the file. With some searching, i found that the channel types that miss in either `scalings` or `unit_scalings` dicts are:
`csd , stim , resp , exci , ias , syst , seeg , dbs , bio , chpi , dipole , gof , ecog , hbo , hbr , temperature , gsr , eyetrack`
This seems like a more general problem to me. I used iteration techniques I saw in the original file, none of which account for `ias` for example. Am I missing something?

drammock:
just a note that I'm about to head on holiday for a week and wasn't able to review this again beforehand. My apologies; hopefully another dev has time to look at it sooner. Otherwise I'll take a look sometime early in December.

Grifunf:
> just a note that I'm about to head on holiday for a week and wasn't able to review this again beforehand. My apologies; hopefully another dev has time to look at it sooner. Otherwise I'll take a look sometime early in December.

No worries, have fun on your holiday! I will continue to work though. Speak soon!

Grifunf:
Hello, writing a small report.
So as of now, I have probably finished [steps 1-2](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1712138949) of the 5 steps before-mentioned and am moving on to the next ones. I have altered quite a few things since the last time we spoke, so I expect to have a lot of corrections - suggestions and problems, and will solve them asap after they are mentioned. Two points among others that definitely need checking are:

- The error of the type 'ias', mentioned in a [comment](https://github.com/mne-tools/mne-qt-browser/pull/212#issuecomment-1814373122) earlier.
- The [disconnect errors](https://github.com/mne-tools/mne-qt-browser/actions/runs/6890458520/job/18743544854#step:11:601), found in previous checks. Right now the checks aren't running, but I think I solved the problem in [this commit](https://github.com/mne-tools/mne-qt-browser/pull/212/commits/c2f02dfbba548a9d33f431ebad1ef68162a60eaf) using the `allow_error` argument.

Now moving on to sensitivity, my thoughts and plans:

1. I am going to check on the auto monitor calibration. I have seen some relevant snippets in the file ([example](https://github.com/mne-tools/mne-qt-browser/blob/fdc9ac77617873655692c2567d6b51ab115e70ba/mne_qt_browser/_pg_figure.py#L3032)), and you have also shared [some useful links](https://github.com/mne-tools/mne-qt-browser/pull/208#issuecomment-1789354644) in the past. Will definitely look into them.
2. But, after discussion with my peers here, it seems that the important enhancement to introduce is the manual monitor calibration, as shown [here](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1727283062) by cbrnr, and as used in EDFbrowser. Among other reasons is that the automatic calibration isn't always reliable, especially when working with VMs or Dockers. So this is probably my main goal.

PR #217, opened 2023-11-30
Remove version file

marsipu:
### What does this implement/fix?
This changes the version-system and adds update-instructions according to mne-tools/mne-python#12245


marsipu:
Tests for python=3.8 seems to have been discontinued by mne-python, right?

hoechenberger (approved):

hoechenberger:
No, it will just use the latest tag as the reference point / version

PR #233, opened 2024-03-04
BUG: Fix bug with multiple theme paths

larsoner:
Closes #231

@cbrnr no idea why #231 doesn't work but at the end of the day here we just want to get an icon from a path... so it's easy enough just to point Qt to the path I think.


cbrnr:
For me, this doesn't fix the problem. When I open the browser from MNELAB (which already uses the .qrc), MNELAB icons disappear.

Also, icons do not adapt to light/dark mode changes on the fly, but that did not work before either.

larsoner:
While on this PR I'd suggest to (make absolutely sure you're on an editable install and then) add breakpoints and/or comment out code and step through the browser init to see when your icons disappear. On this branch I don't think there should be any global theme stuff going on so it should be due to something else

larsoner:
... in MNE-Python itself there is this which is probably still problematic:

https://github.com/mne-tools/mne-python/blob/17d0c87083847f1279b5bd4a941f5daeff40a693/mne/viz/backends/_utils.py#L93

from

https://github.com/mne-tools/mne-python/blob/17d0c87083847f1279b5bd4a941f5daeff40a693/mne/viz/backends/_utils.py#L181

and 

https://github.com/mne-tools/mne-qt-browser/blob/b2f9216a93e4219747044c1efdb52abf95a2c30b/mne_qt_browser/_pg_figure.py#L5176

So I think we'll need similar changes in MNE-Python to fix this. Can you try commenting out this theme search path stuff in MNE-Python to see if it fixes the problem?

larsoner:
... perhaps a clean way to handle this is:

1. Do keep the `setThemeSearchPaths` but prepend the new path rather than replace it
2. Use `QIcon.fromTheme(name, QIcon(fallback_path))` in MNE-Qt-browser and MNE-Python
3. Standardize MNE-Qt-Browser, MNE-Python, and MNELAB around the theme names being "light" and "dark" (ideally but not totally necessary)

This should make it so other stuff is not broken. Then if another app kills the theme search paths (like we have been doing!) then it'll go back to the fallback icon, which could just be the "light" theme version. And if people really want to theme MNE they can do it by providing all "mne standard" names (like "annotations" or whatever). I'm not sure exactly how this would work with MNE-Python and MNE-Qt-Browser both providing a `dark` and `light` theme, though, hopefully it would go through the search paths and go through all matching theme names to find the right icon name.

cbrnr:
This sounds great @larsoner! Thanks! I'll test everything in a couple of days, so if you are in a hurry feel free to merge, we can always amend stuff later.

cbrnr:
I've tested this PR together with https://github.com/mne-tools/mne-python/pull/12483, and everything works! 🚀 🔥 I think I'll revert the change in MNELAB and use the same method like in these PRs (and yes, good idea to standardize themes "light" and "dark").

Thanks @larsoner!

PR #236, opened 2024-03-15
Remove fixes for MNE < 1.0

cbrnr:
Setting the app name in the macOS menu bar is an ugly hack I just realized. The issue is that it overwrites any previously set name, so once again, starting MNE-Qt-Browser from MNELAB changes the name from "MNELAB" to "MNE-Python".

Since (1) this is a hack and (2) it is not really necessary to change the name for MNE-Qt-Browser (the default is "Python"), I would like to remove it.

### DIFF
--- a/mne_qt_browser/_fixes.py
+++ b/mne_qt_browser/_fixes.py
@@ -14,8 +14,8 @@ try:
     from pytestqt.exceptions import capture_exceptions
 except ImportError:
     logger.debug(
-        "If pytest-qt is not installed, errors from inside the event loop will be occluded "
-        "and it will be harder to trace back the cause."
+        "If pytest-qt is not installed, errors from inside the event loop will be "
+        "occluded and it will be harder to trace back the cause."
     )
 
     @contextmanager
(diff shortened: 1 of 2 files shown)


mscheltienne:
How is it set on macOS in MNELAB? Could both use the same mechanism and then MNELAB could overwrite previous entries?

cbrnr:
MNELAB uses the same hack. I'm not sure if it is worth adding this complexity, I'd rather not set the name for the browser since it is obviously not meant to be run as a standalone app (but always from MNE or MNELAB).

hoechenberger:
> it is not really necessary to change the name for MNE-Qt-Browser (the default is "Python"),

I don't think it should simply be called "Python", this is not a great UX. Could we check if a non-default name is currently set, and if yes, don't set something new?

cbrnr:
> I don't think it should simply be called "Python", this is not a great UX. Could we check if a non-default name is currently set, and if yes, don't set something new?

OK, I'll check if this is possible.

cbrnr:
Yes, I think this works: https://github.com/mne-tools/mne-python/pull/12499

Instead of fixing this here as well, I think we just remove these shims as @larsoner suggested.

cbrnr:
OK, this PR now removes all shims required for MNE < 1.0. Since MNE-Qt-Browser already requires MNE >= 1.0, we don't need them anyway.

larsoner:
Thanks @cbrnr

Issue #270, opened 2024-07-10
BUG: MNE_BROWSER_PRECOMPUTE=true broken


larsoner:
On my system I have a lot more memory compared to size of files that I open, so for almost all files precomputation runs.

In theory this is a good thing, but in practice I've noticed some weird bugs where the display doesn't always update properly when I click and unclick bad channels, or enable/disable projectors.

I think relatedly, in #268 the updates only happened properly when I had `MNE_BROWSER_PRECOMPUTE=false` set. Otherwise they didn't take effect. Interestingly, the traces *were* updated when hitting plus or minus on the keyboard. This is maybe some hint for how things should be changed/fixed.

But in any case, we should probably do a comprehensive check of visual output with precompute mode on and off, for at least the following behaviors:

1. Enabling/disabling projectors
2. Clicking/unclicking bad channels that are in projectors when projection is on (as this changes the projector that is applied)
3. Changing channel scalings using plus/minus on the keyboard
4. Changing channel scalings using the new channel-specific UI

@nmarkowitz for now you don't need to worry about any of this stuff -- I just need to remember to test with `MNE_BROWSER_PRECOMPUTE=false` (which I'll do by setting my system config value now).

larsoner:
Whitened channels and maybe stim channels are scaled differently in `_process_data` in MNE, too, so we need to make sure we treat them properly here.

larsoner:
Wow that's quite wrong! 😱 

Yes we should try to fix this precompute issue soon if possible

marsipu:
@larsoner I noticed, that when precompute=True on selecting bad-channels, each time _apply_update_projectors is called and the data is precomputed again. Is that expected or is it related to this issue?

larsoner:
Yep, if the bad channel selection changes the projector should be updated and reapplied!

PR #271, opened 2024-07-10
BUG: Fix bug with over-updating and antialias toggle

larsoner:
In finding #270 I noticed that just opening the Settings dialog triggered 3 updates. There is no reason this is necessary, and is an unnecessary slowdown. It happens because we connect the listener then set the value (to it's currently used value)... which triggers an update.

And this is a bug for `antialiasing` actually because it triggers an AA toggle, so if you had AA on, it connected the listener, set the `setChecked(True)` which initiated a toggle, *disabling* AA. So the check state was always out of phase with the actual state.


larsoner:
@nmarkowitz feel free to take a look and click "approve" if you're happy and I'll merge

larsoner:
This is a pretty trivial change so marking for merge-when-green after merging `main` into this branch

Issue #248, opened 2024-05-01
ENH: Spectrogram viewer


alexrockhill:
### Describe the problem

What would you think about implementing something like this:
![image](https://github.com/mne-tools/mne-qt-browser/assets/13473576/5ad00ad8-a7b3-42a4-9040-7330c801123e)

### Describe your solution

You would probably want to pass frequencies and compute spectrograms on the fly because it would be too much to store in memory

### Describe possible alternatives

You could precompute and decimate. Maybe it would make sense to have both as options.

### Additional context

Any idea how hard this would be? I was thinking I might try and do this in a couple days but if it's going to take a long time maybe it would be better as a GSoC or something.

larsoner:
It is a subgoal listed in https://github.com/mne-tools/mne-python/wiki/GSoC-Ideas#1a-improve-raw-data-browsing-pyqtgraph . @nmarkowitz is going to be doing GSoC to improve mne-qt-browser (yay!) but I don't think this was in the proposal so someone could probably work on it separately.

alexrockhill:
Awesome! I'd be up for giving it a shot. Do you have an idea for the parameter updating interface and what it would look like? To my knowledge that would be a new menu.

larsoner:
We currently have the ability to switch between individual traces and butterfly mode (traces overlaid basically). So I think a reasonable option would be a drop-down that is `Traces / Butterfly / Spectrogram` or similar.

alexrockhill:
Sure but then would that spawn a new menu for changing nfft etc. or use something existing?

larsoner:
It could un-hide or add some toolbar elements for example. Or maybe we want a separate menu. These things we'd need to discuss and converge on, probably easiest to do it with a video meeting at some point. Lots of options / considerations / possible opinions on UI design

nmarkowitz:
My thinking/plan for this is roughly:

- Run a bunch of analyses/preprocessing steps you'd like visualized (like time-frequency or change of reference)
- pass all those datastructures into a call to databrowser
- switch between those with a dropdown selection

My thinking for this approach is that it allows for more complex analyses, like output of a neural network, to be visualized in conjunction with other data. So, hopefully by the end of the summer, this could be incorporated if additional processing is done before databrowser is called

alexrockhill:
[...]

PR #272, opened 2024-07-10
ENH, WIP: Spectrogram mode, minimal version

alexrockhill:
Fixes https://github.com/mne-tools/mne-qt-browser/issues/248.

Hey look, I have a minimally working version and didn't take that long once I got around to this.

[Screencast from 07-10-2024 04:52:50 PM.webm](https://github.com/mne-tools/mne-qt-browser/assets/13473576/a1d94387-c549-49b2-b6b0-3a0f0b8ffbe4)

This isn't the all-in-one data analysis that @nmarkowitz was suggesting, it's just a different mode to view the data in for quality check purposes and if no one reads the help menu and knows to press "f", the browser won't be effected. How it works is that the TFR data is computed on the fly only for the times in view. Will add tests and clean up etc. but it would be nice to get feedback if anyone thinks of things. For instance, there is no way to tell the frequencies at the moment which seems like a bit of an issue. Also, might run slow on a less powerful computer but you can also limit your selection to a smaller time and channel range.

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -38,7 +38,9 @@ from matplotlib.backends.backend_qtagg import FigureCanvasQTAgg
 from matplotlib.colors import to_rgba_array
 from mne import channel_indices_by_type
 from mne.annotations import _sync_onset
+from mne.baseline import rescale
 from mne.io.pick import _DATA_CH_TYPES_ORDER_DEFAULT, _DATA_CH_TYPES_SPLIT
+from mne.time_frequency import tfr_array_morlet
 from mne.utils import _check_option, _to_rgb, get_config, logger, sizeof_fmt, warn
 from mne.viz import plot_sensors
 from mne.viz._figure import BrowserBase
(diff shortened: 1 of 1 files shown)


alexrockhill:
- [ ] Showing more channels results in traces instead of spectrograms

larsoner:
I would say at a minimum you'd need some way to specify / change the frequency range and the colormap limits. Seems like maybe a new dedicated TFR / spectrogram dialog might be needed, or if it's just a few options more rows in the existing Settings dialog could work

larsoner:
> Maybe we should add to remove channel scaling/sensitivity when in spectrogram mode? Probably isn't useful in this scenario.

Yeah it probably does make sense to show/hide based on viewing mode at some point. For now though I think it's okay to keep adding stuff and we can decide later what makes the most sense to show when, or whether to use tabs or different dialogs, etc. Make it work then make it pretty :)

alexrockhill:
Just floating this out: how about `+/-` to zoom in/out on the colormap which uses existing functionality and then `shift+f` to pop up a `fmin/fmax` dialog?

larsoner:
`+/-` makes sense since when not in traces mode it shouldn't matter. But for now rather than a separate dialog I'd put everything in settings then later we can decide the best places to split things out, or use tabs, or show/hide, etc.

alexrockhill:
image

Here's how it looks now

PR #230, opened 2024-02-24
ENH: Monitor Calibration, Sensitivity & Custom Scaling

Grifunf:
#### Reference issue

Resolves mne-tools/mne-python#10888

Previous PRs: mne-tools/mne-qt-browser#208, mne-tools/mne-qt-browser#212

#### What does this implement/fix?

1. **Monitor Calibration**: The user makes two measurements for the width and length of the plot, which are later used for on monitor length calculation and sensitivity.
2. **Sensitivity**: Amplitude unit per length unit, for example `uV` as the amplitude unit and `uV/mm` as the sensitivity unit of measurement for a channel type.
3. **Custom Scaling**: Scaling extended from simple +/- buttons, to being able to specify the Amplitude/Sensitivity for each channel type, or specify the duration, seconds/mm and mm/seconds for the time axis scaling.

#### Information

Created 3 new classes/dialogs to fulfill the above requirements:

1. `CalibrationDialog` for monitor calibration by the user:

calibration

Description: [Inspired by EDFbrowser](https://github.com/mne-tools/mne-python/issues/10888#issuecomment-1727283062). the plot turns in the inverse of the current color, and the user is asked to measure its height and width. Those are saved in `mne.height` and `mne.width` for later use for sensitivity.
In order to keep measurements correct, and not being able to account for window resizes, for the time being the window and plot are locked in size while the user has the calibration mode enabled. While this functionality is enabled the user has access to all sensitivity features stated below and the sensitivity values are displayed alongside the amplitudes..

2. `ScalingDialog` for scaling the amplitude and sensitivity of channels:

scaling off
scaling on

Description: A simple dialog with a row for each channel type, and two columns of text boxes: one for the amplitude, and one for sensitivity. The new added features is being able to scale a specific channel type as opposed to all data together, and also the sensitivity feature. Sensitivity is defined as amplitude unit per mm, on the user screen. Available only after calibration.

3. `TimeScalingDialog` for time axis scaling:

time scaling off

Description: A simple dialog with 3 textboxes. facilitating the seconds/page (duration), the seconds/mm and the mm/second values. Opposed to the previous dialog, the time scaling occurs for every channel type at once. Again, the values relevant to on-screen lengths are available only after calibration.

#### Additional information

1. The code is somewhat well commented, especially where calculations have been made. I am of course available for clarifications.
2. Swapped the old `mne.scale_factor` variable for `mne.scale_factors` dict in order to scale differently each channel type.
[...]

PR #273, opened 2024-07-11
Enh: channel sensitivity ui

nmarkowitz:
#### Reference issue

https://github.com/mne-tools/mne-qt-browser/pull/230 , https://github.com/mne-tools/mne-qt-browser/pull/212 , https://github.com/mne-tools/mne-qt-browser/pull/208 , https://github.com/mne-tools/mne-python/issues/10888

Adds widget to settings dialog displaying units of sensor in relation to screen size. Ability to edit widgets will be done in future pull request

Screenshot 2024-07-10 at 8 38 05 PM

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -4164,6 +4164,9 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             self.mne.ax_vscroll.update_nchan()
             self.mne.plt.setYRange(ymin, ymax, padding=0)
 
+        if self.mne.fig_settings is not None:
+            self.mne.fig_settings._update_sensitivity_spinbox_values()
+
     def _remove_vline(self):
         if self.mne.vline is not None:
             if self.mne.is_epochs:
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1891,7 +1891,7 @@ class SettingsDialog(_BaseDialog):
             inv_norm = _get_channel_scaling(self, ch_type)
             ch_scale_spinbox.setValue(inv_norm)
             ch_scale_spinbox.valueChanged.connect(
-                _methpartial(self._update_spinbox_values, ch_type=ch_type)
+                _methpartial(self._update_scaling_spinbox_values, ch_type=ch_type)
             )
             self.ch_scaling_spinboxes[ch_type] = ch_scale_spinbox
 
@@ -1945,7 +1945,7 @@ class SettingsDialog(_BaseDialog):
     def _toggle_antialiasing(self, _):
         self.weakmain()._toggle_antialiasing()
 
-    def _update_spinbox_values(self, *args, **kwargs):
+    def _update_scaling_spinbox_values(self, *args, **kwargs):
         """Update spinbox values. If any args or kwargs do a specific channel update."""
         # If new value for a channel given update that channel type and redraw
         if len(args) > 0:
@@ -4044,7 +4044,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         # Update Scalebars
         self._update_scalebar_values()
         if self.mne.fig_settings is not None:
-            self.mne.fig_settings._update_spinbox_values()
+            self.mne.fig_settings._update_scaling_spinbox_values()
 
     def hscroll(self, step):
         """Scroll horizontally by step."""
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1955,7 +1955,7 @@ class SettingsDialog(_BaseDialog):
             # If new_value is 0 then scaling is stuck on 0.
             # To get out of 0 set scalings to 1 and then set the new value
             if new_value == 0:
-                self.mne.scalings[ch_type] = 0
+                self.mne.scalings[ch_type] = 1e-12
             else:
                 self.mne.scalings[ch_type] = 1
                 self.mne.scalings[ch_type] = new_value / _get_channel_scaling(
@@ -4004,7 +4004,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
 
     def _update_ch_spinbox_values(self):
         if self.mne.fig_settings is not None:
-            self.mne.fig_settings.update_all_spinboxes()
+            self.mne.fig_settings._update_scaling_spinbox_values()
 
     def _set_scalebars_visible(self, visible):
         for scalebar in self.mne.scalebars.values():
@@ -4853,7 +4853,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         self.mne.butterfly = butterfly
         self._update_picks()
         self._update_data()
-        self._update_ch_spinbox_values()
+        # self._update_ch_spinbox_values()
 
         if butterfly and self.mne.fig_selection is not None:
             self.mne.selection_ypos_dict.clear()
@@ -4882,6 +4882,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         if self.mne.fig_selection is not None:
             # Update Selection-Dialog
             self.mne.fig_selection._style_butterfly()
+            self._update_ch_spinbox_values()
 
         # Set vertical scrollbar visible
         self.mne.ax_vscroll.setVisible(
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2003,9 +2003,9 @@ class SettingsDialog(_BaseDialog):
             self.ch_sensitivity_spinboxes[ch_type].setValue(
                 _calc_chan_type_to_physical(self, ch_type, units=current_units)
             )
-            self.ch_sensitivity_spinbox_labels[ch_type].setText(
-                f"{ch_type} ({self.mne.units[ch_type]}/{current_units})"
-            )
+            # self.ch_sensitivity_spinbox_labels[ch_type].setText(
+            #    f"{ch_type} ({self.mne.units[ch_type]}/{current_units})"
+            # )
 
 
 class HelpDialog(_BaseDialog):
@@ -5335,6 +5335,11 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         # Make sure it gets deleted after it was closed.
         self.deleteLater()
         self._closed = True
+
+    def resizeEvent(self, event):
+        super().resizeEvent(event)
+        if self.mne.fig_settings is not None:
+            self.mne.fig_settings._update_sensitivity_spinbox_values()
 
     def _fake_click_on_toolbar_action(self, action_name, wait_after=500):
         """Trigger event associated with action 'action_name' in toolbar."""
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -1866,22 +1866,24 @@ class SettingsDialog(_BaseDialog):
         self.scroll_sensitivity_slider.setValue(self.mne.scroll_sensitivity)
         layout.addRow("horizontal scroll sensitivity", self.scroll_sensitivity_slider)
 
-        # Add subgroup box to show channel type scalings
         layout.addItem(QSpacerItem(10, 10, QSizePolicy.Minimum, QSizePolicy.Expanding))
-        ch_scaling_box = QGroupBox("Channel Type Scalings")
-        ch_scaling_box.setStyleSheet("QGroupBox { font-size: 12pt; }")
-        ch_scaling_layout = QFormLayout()
-        self.ch_scaling_spinboxes = {}
 
-        # Get all unique channel types and allow scaling
+        # Get all unique channel types
         ordered_types = self.mne.ch_types[self.mne.ch_order]
         unique_type_idxs = np.unique(ordered_types, return_index=True)[1]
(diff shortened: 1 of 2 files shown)


larsoner:
Rather than another set of rows with the channel types duplicated it probably makes more sense as a grid layout like

| Type |Scaling|Sensitivity |
| -- | -- | -- |
| grad (fT/cm) | 800 ↕ | 111 ↕ |
| mag (fT) | 2000 ↕ | 277.5 ↕ |
| eeg (uV) | 40 ↕ | 5.5 ↕ |
| eog (uV) | 300 ↕ | 41.6 ↕ |
| |  | / mm ⌄ |

nmarkowitz:
> Rather than another set of rows with the channel types duplicated it probably makes more sense as a grid layout like
> 
> Type	Scaling	Sensitivity
> grad (fT/cm)	800 ↕	111 ↕
> mag (fT)	2000 ↕	277.5 ↕
> eeg (uV)	40 ↕	5.5 ↕
> eog (uV)	300 ↕	41.6 ↕
> / mm ⌄

I like this idea. Just pushed it

larsoner:
> I like this idea. Just pushed it

Well not quite... you still have redundant channel type names and units (other than the "per distance") in there, and they don't line up in rows either:

image

In other words, you did something like this rather than what I suggested:

| Scalings | | Sensitivities | |
| -- | -- | -- | -- |
| grad (fT/cm) | 800 | grad (fT/cm/mm) | 93.6 |

I think my suggestion is more compact / consistent but contains all the same information. Can you see if my suggestion works?

By eye looking at the `cm` it appeared to be correct, but you'll need to now also connect it to the resize event so that the values update when you resize. (The way I wanted to check the correctness was to resize my window such that the scale bar was very close to 1cm then check that the scale bar matched the `Sensitivities` value, but the spinbox values didn't update when I changed my window size.)

nmarkowitz:
How does this look now? This is what was is your original suggestion. I'm wondering if the dropdown should be placed elsewhere. And for the next PR we're probably gonna need some toggle button like a checkbox or dropdown to choose which units are displayed on the scalebars
Screenshot 2024-07-12 at 8 00 23 PM

mne_qt_browser/_pg_figure.py
```diff
@@ -230,6 +230,63 @@ def _get_channel_scaling(widget, ch_type):
     return inv_norm
 
 
+def _calc_chan_type_to_physical(widget, ch_type, units="mm"):
+    """Convert data to physical units."""
+    # Load in some data and measure its range in a timeperiod
+    start, stop = widget.weakmain()._get_start_stop()
+    data, _ = widget.weakmain()._load_data(start, stop)
```
larsoner, line 237: @nmarkowitz I resized my window to show 800 (fT/cm)/cm so that the MEG scale bar should be 1cm but it was a bit big. Adjusting the `__init__` of `class ScaleBar` to use a flat instead of rounded cap (to not draw farther than we should) seemed to fix it:
```
        pen = self.mne.mkPen(color="#AA3377", width=5)
        pen.setCapStyle(Qt.FlatCap)
        self.setPen(pen)
```

However, you should not need to know anything about the actual *data values* to know the effective scale on screen. So using `data` at all (and derivatives like `peak_to_peak_data`) like you do here should not be needed. The only thing that should matter are:

1. The factors that are used to scale the data
2. The number of pixels used to / allocated to show those data (which you should get from the viewbox somehow, could be correctly below?), which you can get from the proportion of the viewbox (in pixels) that is dedicated to showing the data

From those two things you should be able to figure out the effective scaling. Unpacking your code later you have (for `mm` at least):
```
        pixels_per_data_unit = pixel_height / data_height
        peak_to_peak_pixels = peak_to_peak_data * pixels_per_data_unit
        dpi = QApplication.primaryScreen().logicalDotsPerInch()
        peak_to_peak_inches = peak_to_peak_pixels / dpi
        peak_to_peak_mm = peak_to_peak_inches * 25.4
        if units == "mm":
            return (
                _get_channel_scaling(widget, ch_type)
                * peak_to_peak_data
                / peak_to_peak_mm
```
which is equivalent to
```
        dpi = QApplication.primaryScreen().logicalDotsPerInch()
        peak_to_peak_mm = peak_to_peak_data * pixel_height / data_height / dpi * 25.4
        if units == "mm":
            return (
                _get_channel_scaling(widget, ch_type)
                * peak_to_peak_data
                / peak_to_peak_mm
            )
```
which in turn is equivalent to
```
        dpi = QApplication.primaryScreen().logicalDotsPerInch()
        if units == "mm":
            return (
                _get_channel_scaling(widget, ch_type)
                / (pixel_height / data_height / dpi * 25.4)
            )
```
I'll push a commit to simplify the calculation once I verify it's correct

mne_qt_browser/_pg_figure.py
```diff
@@ -1809,34 +1842,86 @@ def __init__(self, main, title="Settings", **kwargs):
...
-                ch_spinbox.setValue(inv_norm)
-                ch_spinbox.valueChanged.connect(
-                    _methpartial(self._update_spinbox_values, ch_type=ch)
-                )
-                self.ch_scaling_spinboxes[ch] = ch_spinbox
-                ch_scaling_layout.addRow(f"{ch} ({self.mne.units[ch]})", ch_spinbox)
+        ch_types_ordered = [
+            ordered_types[idx]
+            for idx in sorted(unique_type_idxs)
+            if ordered_types[idx] in self.mne.unit_scalings.keys()
```
larsoner, line 1853: For `dict` you can directly query the dict itself for membership rather than checking for membership in the `.keys()`
```suggestion
            if ordered_types[idx] in self.mne.unit_scalings
```

mne_qt_browser/_pg_figure.py
```diff
@@ -1809,34 +1842,86 @@ def __init__(self, main, title="Settings", **kwargs):
...
+            ch_sens_spinbox.setStepType(QAbstractSpinBox.AdaptiveDecimalStepType)
+            ch_sens_spinbox.setReadOnly(True)
+            ch_sens_spinbox.setDisabled(True)
+            ch_sens_spinbox.setValue(
+                _calc_chan_type_to_physical(self, ch_type, units=current_units)
+            )
+            self.ch_sensitivity_spinboxes[ch_type] = ch_sens_spinbox
+            # self.ch_sensitivity_spinbox_labels[ch_type] = QLabel(
+            #     f"{ch_type} ({self.mne.units[ch_type]}/{current_units})"
+            # )
```
larsoner, line 1912: Remove cruft

larsoner (commented):
Looks good, pushed a commit!

As usual, next I think some tests would be good. For example you could just get whatever the current scalings values are, change the units to `/ cm` or something and then query the double spin box values and make sure they changed by the correct amount. You could also resize the window by 10% for example and make sure the values change by 10%.

mne_qt_browser/_pg_figure.py
nmarkowitz: Thanks for doing this. I think's it's ready to be pulled

larsoner:
I just did:
```
mne browse_raw ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif
```
and then opened the settings dialog and pressed `"b"` to check to see if the values updated properly when butterfly mode was activated (because this changes the scalebar size). Have you checked? I get an error:
```
$ mne browse_raw ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif
  File "/home/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 5071, in keyPressEvent
    slot()
  File "/home/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 4903, in _toggle_butterfly
    self._set_butterfly(not self.mne.butterfly)
  File "/home/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 4856, in _set_butterfly
    self._update_ch_spinbox_values()
  File "/home/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 4007, in _update_ch_spinbox_values
    self.mne.fig_settings.update_all_spinboxes()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SettingsDialog' object has no attribute 'update_all_spinboxes'
```
Assuming you can replicate, once you fix the bug you can to make sure the values update properly when activating/deactivating butterfly mode. But keep in mind https://github.com/mne-tools/mne-qt-browser/issues/276 where I point out that butterfly mode is already perhaps somewhat broken. It might be hard for you even to measure the size of the scale bar given the overlap... but you can at least eyeball the top half of the scale bar for the topmost channel type and go based on that I think.

nmarkowitz:
Thanks for catching that. Just pushed commit

nmarkowitz:
@larsoner every failed test is due to the sensitivity unit test I added. It passes on my computer but I don't know how to make it pass on github. Maybe I should change what the unit test is?

larsoner:
It could be some little precision issue, I think it's okay to use `assert_allclose(a, b, atol=0.1)` to tolerate the rounded term being different

nmarkowitz:
Just pushed using `assert_allclose()` for unit test. I set tolerance for 2.

mne_qt_browser/tests/test_pg_specific.py
```diff
@@ -295,6 +295,20 @@ def test_pg_settings_dialog(raw_orig, pg_backend):
...
+    scaling_spinbox_value = ch_spinbox.value()
+    sensitivity_spinbox_value = ch_sens_spinbox.value()
+    scaling_value = fig.mne.scalings[ch_type_test]
+    new_scaling_spinbox_value = scaling_spinbox_value * 2
+    new_expected_sensitivity_spinbox_value = sensitivity_spinbox_value * 2
+    ch_scale_spinbox.setValue(new_scaling_spinbox_value)
+    new_scaling_value = fig.mne.scalings[ch_type_test]
+    assert scaling_value != new_scaling_value
+    np.testing.assert_allclose(
+        ch_sens_spinbox.value(), new_expected_sensitivity_spinbox_value, atol=2
```
larsoner, line 309: Why 2? It seems like `atol=0.1` (or maybe 0.2) should be enough?

larsoner: ... let's see if CIs agree
```suggestion
        ch_sens_spinbox.value(), new_expected_sensitivity_spinbox_value, atol=0.1
```

larsoner:
1. Did you measure to check the values? When I plot set the window size to match the stated sensitivity in inches:

   image

   The physical measurement shows closer to 1/4" from the top of the "grad" scale bar to where the "grad" text is, suggesting that its full size is 1/2" for the 800 fT/cm. So I think there's a factor of two error?

2. From this view, when I hit "b" the plot updates but the spinbox values don't update:

   image

   If I trivially resize the window e.g. in the horizontal direction, they do:

   image

   So I think the "spin box update when butterfly mode is toggled" still needs to be updated / connected.

nmarkowitz:
I fixed the butterfly toggle. 

For the butterfly sensitivity I think it's accurate. It's difficult to tell with overlapping scalebars. I tried measuring it with just one channel type on the screen. Though it could have been due to a prior error. 

Screenshot 2024-07-22 at 10 50 21 AM

larsoner:
Now I have a bug where if I do:
```
$ mne browse_raw ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif --clipping None --butterfly
```
This one looks correct, the distance from the top of the `grad` bar to the `grad` ylabel tick is very close to 1/4" so the full bar (assuming it's centered) is 1/2", and the scale bar is labeled as 400 fT/cm in the plot and the spin boxes, suggesting the sensitivity is 400 fT/cm/half-inch, which is the same as 800 fT/cm/inch (i.e., the spin box is correct correct):

image

But if I do the same command *without* the `--butterfly` then interactively do `"b"` on the keyboard to toggle it, I get the same plot and same spin box values as before **except text on the scale bar says 800 fT/cm** despite the text in the spin boxes being update:

image

So I think there's some bug with updating the scale bar values perhaps, but I think we can chalk this up to https://github.com/mne-tools/mne-qt-browser/issues/276. Might be a good one to tackle next if you want @nmarkowitz or you can move on to making the sensitivity boxes interactive (I expect this one might be easy now hopefully)!

mne_qt_browser/_pg_figure.py
```diff
@@ -1809,34 +1842,86 @@ def __init__(self, main, title="Settings", **kwargs):
...
```
larsoner, line 1912:
```suggestion
```

mne_qt_browser/_pg_figure.py
```diff
@@ -4758,7 +4856,7 @@ def _set_butterfly(self, butterfly):
         self.mne.butterfly = butterfly
         self._update_picks()
         self._update_data()
-        self._update_ch_spinbox_values()
+        # self._update_ch_spinbox_values()
```
larsoner, line 4859:
```suggestion
```

mne_qt_browser/_pg_figure.py
```diff
@@ -4787,6 +4885,7 @@ def _set_butterfly(self, butterfly):
         if self.mne.fig_selection is not None:
             # Update Selection-Dialog
             self.mne.fig_selection._style_butterfly()
+            # self._update_ch_spinbox_values()
```
larsoner, line 4888:
```suggestion
```

larsoner (commented):
Just a few cruft cleanups which I'll commit then mark for merge-when-green, thanks in advance @nmarkowitz !

FYI to spot these things, after you think your code is good for review and you push, I typically try to add one more step before pinging for re-review -- looking briefly at the diff on GitHub in the "Files" tab like https://github.com/mne-tools/mne-qt-browser/pull/273/files. A quick 30-sec scan of your own code probably would have caught these leftover lines to be removed!

PR #278, opened 2024-07-22
Enh sensitivity UI change

nmarkowitz:
#### Reference issue

https://github.com/mne-tools/mne-qt-browser/pull/230 , https://github.com/mne-tools/mne-qt-browser/pull/212 , https://github.com/mne-tools/mne-qt-browser/pull/208 , https://github.com/mne-tools/mne-python/issues/10888 , https://github.com/mne-tools/mne-qt-browser/pull/273

Extends  https://github.com/mne-tools/mne-qt-browser/pull/273 by enabling sensitivity spinboxes to be updated

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -2043,7 +2043,7 @@ class SettingsDialog(_BaseDialog):
             mon_height_inch = _convert_physical_units(
                 new_ht_val, from_unit=mon_units, to_unit="inch"
             )
-            dpi = (px_height / dpr) / mon_height_inch
+            dpi = (px_height / mon_height_inch) / dpr
 
             # Find new width of monitor
             with SignalBlocker(self.mon_width_spinbox):
@@ -2069,7 +2069,7 @@ class SettingsDialog(_BaseDialog):
             mon_width_inch = _convert_physical_units(
                 new_wd_value, from_unit=mon_units, to_unit="inch"
             )
-            dpi = (px_width / dpr) / mon_width_inch
+            dpi = (px_width / mon_width_inch) / dpr
 
             # Find new height of monitor
             with SignalBlocker(self.mon_height_spinbox):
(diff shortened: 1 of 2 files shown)


nmarkowitz:
@larsoner can you remind me, how can I calculate logical dpi and where should `QApplication.primaryScreen().devicePixelRatio()` be used for this calculation?

larsoner:
... or rather, the wrong stuff updates when I click the `/ inch` in the monitor section. It updates the *upper* set of boxes, when really it should update the *lower* set. In other words if I have this to start:

Screenshot 2024-08-07 at 11 32 00 AM

If I click the `Monitor Size` dropdown `/ mm` and change it to `/ inch`, the *sensitivity* values shouldn't change, the *monitor* values should change to `222.52mm/25.4=8.76in`, i.e., it changes the units but not the screen size, it doesn't use the old `float` with the new units to set a new size. So what it does now, which is this, is incorrect:

Screenshot 2024-08-07 at 11 32 28 AM

It should be more like this:

Screenshot 2024-08-07 at 11 33 46 AM

But even when I change it manually to a height of 8.76 (as in the image above) something is wrong, because the `grad` sensitivity is now `73.6 fT/cm / mm` instead of `83.1 fT/cm / mm` despite inputting the same physical size for my monitor (granted, in different units, but this should be accounted for properly) and not changing the window size.

And regarding the "no update on resize", I noticed that when I resize I do get this in the terminal:

```
  File "/Users/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py", line 5544, in resizeEvent
    self.mne.fig_settings._update_sensitivity_spinbox_values()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SettingsDialog' object has no attribute '_update_sensitivity_spinbox_values'
```

mscheltienne:
FYI: https://mne.discourse.group/t/how-access-small-vertical-voltage-line-when-plotting-raw-eeg-changing-voltage-amplitude-when-plotting/9243

nmarkowitz:
@mscheltienne @larsoner I think this is ready. Can you check out and confirm for yourselves?

wmvanvliet:
Functionally, this works very nicely on my Windows machine.

larsoner:
@nmarkowitz I assume you just forgot to take the PR out of draft mode? Works well for me as well so I'll go ahead and merge, thanks!

PR #292, opened 2024-10-21
ENH: Swallow PySide6 warning

larsoner:
Get rid of this unhelpful warning:
```
/home/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py:3575: RuntimeWarning: Failed to disconnect (None) from signal "triggered()".
  sig.disconnect()
```
And also test against latest dev branch of pyqtgraph.


larsoner:
Self-merging to keep CIs green and this seems fairly trivial

PR #308, opened 2025-01-14
Fix annotation of last sample

richardkoehler:
#### Reference issue

Fixes issue #12893 in mne-python (https://github.com/mne-tools/mne-python/issues/12893)
The issue is also discussed in PR #12895 in mne-python (https://github.com/mne-tools/mne-python/pull/12895)

Currently, when annotating bad epochs in mne-qt-browser, annotations were limited to the last sample of the recording, so that the last sample of the recording would not actually be rejected.
This fixes this issue, and additionally allows for 1-sample annotations at the start (t=0) and end (t=xmax) of the recording.

- raise upper limit of RawViewBox to xmax + 1 sample, so that new annotations can extend above the recording
- raise upper limit of AnnotRegion to xmax + 1 sample for all annotations
- to allow for 1-sample annotations at the beginning (0) and end (xmax) of recording in the AnnotationDock :
    - raise maximum of an annotation start to xmax
    - lower the minimum of an annotation end/stop to 0
    - raise maximum of an annotation end to xmax + 1 sample

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -3261,7 +3261,7 @@ class AnnotationDock(QDockWidget):
         start = self.start_bx.value()
         sel_region = self.mne.selected_region
         stop = sel_region.getRegion()[1]
-        if start < stop:
+        if start <= stop:
             self.mne.selected_region.setRegion((start, stop))
             # Make channel specific fillBetweens stay in sync with annot region
             # if len(sel_region.single_channel_annots.keys()) > 0:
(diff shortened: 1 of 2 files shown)


larsoner:
@richardkoehler can you look at modifying some test such that it would fail on `main` but pass on this PR? We have some tests that (via code) click-and-drag annotations I think, so I think you should be able to follow those and adapt appropriately

richardkoehler:
@larsoner Sure, it took me a while to get the hang of the simulated clicks and drags, but I managed now. I implemented two new tests that pass here but fail in main. One other test also fails in main, but passes in mine, since it was checking start<stop (and vice versa), which we changed to <=.
I downsampled the testing data in my tests, which is not strictly necessary, but speeds them up.

larsoner (approved):
Beautiful, thanks @richardkoehler !

richardkoehler:
No problem, thanks for the support and the merge!

PR #314, opened 2025-02-25
Enhance Help Menu for Better Functionality Discoverability

### LINKED ISSUE #84: Add a proper menu to make functionality more discoverable
In https://github.com/mne-tools/mne-qt-browser/issues/79 it became apparent that not even very sophisticated and knowledgeable users like @agramfort are aware of all the cool functionality baked into this tool. This is partially because these features are difficult to discover and only usable via keyboard shortcuts, for which we provide a rather poor table after clicking the Help button. 

I propose we add a proper menu that makes for easier discoverability of both the functionality and their associated keyboard shortcuts (as they appear next to the menu entries on most (all?) systems).

mohitkumhar:
#### Reference issue

Example: Fixes #84 

#### What does this fix?

This PR improves the Help menu by replacing the old HelpDialog with a simpler and more accessible menu. Now, users can find keyboard and mouse shortcuts directly in a dropdown menu instead of a separate dialog.

Key changes
- Removed the HelpDialog class.
- Introduced a Help button (QToolButton) for quick access.

#### Additional information

Please review the changes and let me know if there are any improvements or issues that need to be addressed. 

<h4>Here are the screenshots of the UI</h4>

image

image

### DIFF
--- a/mne_qt_browser/_pg_figure.py
+++ b/mne_qt_browser/_pg_figure.py
@@ -4236,7 +4236,6 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
 
     def create_menus(self):
         """Creates and returns the application's menu bar with properly aligned shortcuts."""
-
         if not self.menuBar():
             self.setMenuBar(QMenuBar(self))
         menu_bar = self.menuBar()
@@ -4272,7 +4271,6 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 ("Toggle Events visibility     ", "E"),
                 ("Toggle Projection Figure / all projections     ", "J"),
             ],
-
             view_menu: [
                 ("Add/remove channels     ", "Shift + Click"),
                 ("Decrease duration (¼ page)     ", "Home"),
@@ -4291,13 +4289,11 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 ("Toggle Crosshair     ", "X"),
                 ("Toggle Zen Mode     ", "Z"),
             ],
-
             help_menu: [
                 ("Show Help     ", "?"),
                 ("Toggle Fullscreen     ", "F11"),
                 ("Close     ", "Escape"),
             ],
-
             scroll_menu: [
                 ("Scroll left (¼ page/full page)     ", "← / →"),
                 ("Scroll up (full page)     ", "↑"),
--- a/mne_qt_browser/tests/test_pg_specific.py
+++ b/mne_qt_browser/tests/test_pg_specific.py
@@ -486,6 +486,7 @@ def test_pg_settings_dialog(raw_orig, pg_backend):
     fig.resize(orig_window_size.width() * 2, orig_window_size.height() * 2)
     assert ch_sens_spinbox.value() != orig_sens
 
+
 def test_pg_toolbar_time_plus_minus(raw_orig, pg_backend):
     """Test time controls."""
     fig = raw_orig.plot()
@@ -650,6 +651,7 @@ def test_pg_toolbar_actions(raw_orig, pg_backend):
     assert pg_backend._get_n_figs() == 3
     fig._fake_click_on_toolbar_action("Settings", wait_after=100)
     assert pg_backend._get_n_figs() == 2
+
 
 # LAB values taken from colorspacious on 2024/06/10
 @pytest.mark.parametrize(


mne_qt_browser/_pg_figure.py
```diff
@@ -2228,86 +2228,88 @@ def _update_spinbox_values(self, *args, **kwargs):
...
-        self.setLayout(layout)
-        self.show()
-
-        # Set minimum width to avoid horizontal scrolling
-        scroll_area.setMinimumWidth(
-            scroll_widget.minimumSizeHint().width()
-            + scroll_area.verticalScrollBar().width()
-        )
-        self.update()
+# Replaced this class with a function declared in MNEQtBrowser
```
larsoner, line 2231: Should be able to delete all this, no need to leave commented. It will live on in our `git history`

larsoner (commented):
Could you post some screenshots of the new UI? That would facilitate review

All PRs should also ideally add some sort of test that hits the new lines. For your PR I guess that would be faking a click in the GUI on the menu and ensuring a new modal dialog window is visible

mne_qt_browser/_pg_figure.py
mohitkumhar: I have added screenshots of both the Keyboard Shortcuts and Mouse Controls dropdown menus.

Currently, the menu opens a dropdown when hovered, but it does not show a separate pop-up (modal) when clicked. Let me know if this behavior meets your requirements or if you need a modal dialog instead.

If no further changes are needed, I will remove the commented-out code as suggested.
Otherwise, I will make all necessary changes and push them together.



image

image

mohitkumhar: Hey, I just removed the comments, you can check it out

larsoner:
@mohitkumhar I think what @hoechenberger had in mind was a top-level menu like that in https://github.com/mne-tools/mne-qt-browser/pull/126. It looks like you added a menu that is accessed via a toolbar button instead

mohitkumhar:
Hey @larsoner  ,
Kindly check this UI , is it the expected output?


Image

Image

hoechenberger:
@mohitkumhar Thanks for your work on this!

What I actually had in mind was adding a full menu bar, though. Something like what you'd see in other applications like e.g. MS Word:

Screenshot 2025-03-07 at 13 23 00

Since menu items can be linked to keyboard shortcuts (which will be displayed in the menu itself), these would be kind of self-documenting.

mohitkumhar:
Sure @hoechenberger I understand it, let me create menu and share sample

hoechenberger:
You can take #126 as a starting point

mohitkumhar:
Hey @hoechenberger , Kindly check it out!

https://github.com/user-attachments/assets/e8451c65-37ec-4e9b-b21b-f8c6f8149135

mohitkumhar:
Hey , kindly check the commit I made

mne_qt_browser/_pg_figure.py
```diff
@@ -4234,6 +4232,80 @@ def __init__(self, **kwargs):
...
+            action.setDefaultWidget(widget)
+            menu.addAction(action)
+
+        shortcut_data = {
+            mne_python: [
+                ("Create annotation     ", "Click + Drag"),
+                ("Mark/unmark bad channel     ", "Left-click channel"),
+                ("Mark/unmark bad segment     ", "Left-click data"),
+                ("Toggle Annotation Tool / Visibility     ", "A"),
+                ("Toggle Butterfly     ", "B"),
```
larsoner, line 4269: Looks like a good start. A few ideas:

1. From a UI perspective, a clickable menu item should initiate an action. Some of these can't, such as "Create Annotation". These should at least be "disabled" (i.e., still visible but not clickable). Not 100% sure we want these in a menu but... maybe okay to leave them.
2. Some of them when clicked should initiate the action, such as the "Toggle DC Correction" -- clicking it should actually toggle DC correction.

Also this code isn't quite DRY... every first option has a bunch of extra spaces at the end. (1) I don't think these should be needed, and (2) if they absolutely are, they could be added to `add_menu_action` rather than being in every one of these entries.

mne_qt_browser/_pg_figure.py
```diff
@@ -4234,6 +4232,80 @@ def __init__(self, **kwargs):
...
+        menu_bar = self.menuBar()
+        mne_python = menu_bar.addMenu("MNE-Python")
+        view_menu = menu_bar.addMenu("View")
+        help_menu = menu_bar.addMenu("Help")
+        scroll_menu = view_menu.addMenu("Scroll")
+
+        def add_menu_action(menu, description, shortcut):
+            widget = QWidget()
+            layout = QHBoxLayout(widget)
+            layout.setContentsMargins(10, 2, 10, 2)
```
larsoner, line 4250: ... this could probably be tweaked rather than using spaces below to fix the spacing.

mne_qt_browser/_pg_figure.py
```diff
@@ -4234,6 +4232,80 @@ def __init__(self, **kwargs):
...
+            scroll_menu: [
+                ("Scroll left (¼ page/full page)     ", "← / →"),
+                ("Scroll up (full page)     ", "↑"),
+                ("Scroll down (full page)     ", "↓"),
+            ],
+        }
+
+        for menu, items in shortcut_data.items():
+            for description, shortcut in items:
+                add_menu_action(menu, description, shortcut)
```
hoechenberger, line 4306: You should use the same shortcuts as defined for the browser window, e.g.,

```python
    help_menu.addAction(
         'Show keyboard shortcuts',
         browser._toggle_help_fig,
         shortcut=browser.mne.keyboard_shortcuts['?']['qt_key']
     )
```

See #126

hoechenberger (changes requested):

PR #320, opened 2025-04-03
ENH: Make _process_data arg/kwarg agnostic

larsoner:
This allows the extra-data-loading gymnastics in https://github.com/mne-tools/mne-python/pull/13183. Also adds a little test that the downsampling stuff actually works (I wasn't sure looking at the code at first).


larsoner:
Contains a fix for the pip-pre failure in #322 but I'll push an empty commit just to make sure

drammock (approved):

Issue #342, opened 2025-07-29
Store Overview Mode setting


cbrnr:
### Describe the problem

Currently, the browser does not remember the Overview Mode setting, i.e., visible or hidden, and the contents if visible (Emtpy or Channels).

BTW I get an error when trying to change the contents, but this is unrelated:

```
Traceback (most recent call last):
  File "mne_qt_browser/_pg_figure.py", line 3570, in call
    return meth_(*args_, **kwargs, **kwargs_)
TypeError: MNEQtBrowser._overview_radio_clicked() missing 1 required positional argument: 'checked'
```

### Describe your solution

I would store the current state in the settings. I'll submit a PR soon (plus another one to fix the error).

### Describe possible alternatives

We could also add these options to the settings dialog, in which case users could customize the defaults. However, I think just storing the current values is easier and better.

PR #343, opened 2025-07-29
Fix Overview Mode toggle

cbrnr:
Fixes a bug where the Overview Mode could not be selected (mentioned in #342).


cbrnr:
This might be specific to `PySide6` – can you try? I do not get an error with `PyQt6` either.

Additional observation: I saw that you get the "z-score" option, but I don't. Do you know why?

mne_qt_browser/_pg_figure.py
```diff
@@ -4318,11 +4323,12 @@ def _toggle_scalebars(self):
 
     def _overview_mode_changed(self, new_mode):
         self.mne.overview_mode = new_mode
+        self._save_setting("overview_mode", new_mode)
         self.mne.overview_bar.set_background()
         if not self.mne.overview_bar.isVisible():
             self._toggle_overview_bar()
 
-    def _overview_radio_clicked(self, checked, *, menu, new_mode):
+    def _overview_radio_clicked(self, checked=False, *, menu, new_mode):
```
larsoner, line 4331: Not surprised this new default is required, I had to do something somewhere else fairly recently for 6.9+ PySide

PR #356, opened 2025-08-08
Use step size of 1 to increase/decrease channels

### LINKED ISSUE #352: Toolbar buttons should increase/decrease number of channels by 1
The two toolbar buttons currently increase and decrease the number of visible channels by 10. This behavior is unexpected and makes difficult to fine-tune the display, so I suggest changing the increment/decrement to 1.

The only potential drawback I see is that increasing the number of visible channels from e.g. 10 to 200 would take longer with smaller steps. However, in such cases, users can use the keyboard shortcuts (Shift + PageUp/PageDown I believe) to adjust the view more efficiently.

cbrnr:
Fixes #352. This PR changes the step size to 1 for the toolbar buttons that change the number of shown channels, making it consistent with the PageUp/PageDown keyboard shortcuts.

### DIFF
--- a/src/mne_qt_browser/_pg_figure.py
+++ b/src/mne_qt_browser/_pg_figure.py
@@ -568,14 +568,14 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             text="Show fewer channels",
             parent=self,
         )
-        adecr_nchan.triggered.connect(_methpartial(self.change_nchan, step=-10))
+        adecr_nchan.triggered.connect(_methpartial(self.change_nchan, step=-1))
         self.mne.toolbar.addAction(adecr_nchan)
         aincr_nchan = QAction(
             icon=self._qicon("more_channels"),
             text="Show more channels",
             parent=self,
         )
-        aincr_nchan.triggered.connect(_methpartial(self.change_nchan, step=10))
+        aincr_nchan.triggered.connect(_methpartial(self.change_nchan, step=1))
         self.mne.toolbar.addAction(aincr_nchan)
         self.mne.toolbar.addSeparator()
 


larsoner:
Failure looks related

cbrnr:
Yep, should be fixed.

Fun fact as an aside: MNE-Python has to be installed in editable mode, otherwise the test file is not available. So the discussion on including the tests with the package is moot, since they wouldn't run anyway.

tests/test_pg_specific.py
```diff
@@ -594,30 +595,28 @@ def test_pg_toolbar_channels_plus_minus(raw_orig, pg_backend):
     if fig.mne.butterfly is True:
         fig._fake_keypress("b")  # toggle butterfly off
 
-    for _ in range(10):
-        if ymax - ymin <= 2:
-            break
+    while ymax - ymin > 2:  # while more than 1 channel visible
```
larsoner, line 598: I actually prefer the old pattern... `while` is dangerous and can cause nasty timeouts (could be like 80 min on CI etc.!) and the old `for` will not. Even better would be:
```
for ...
    if ...
        ...
        break
else:
    raise RuntimeError("Condition not met!")  # pragma: no cover
```

cbrnr: Then we have to bump the number of iterations quite a bit, because 10 were only sufficient if you go in steps of 10. I'm not sure this is really a problem, and to be honest, I don't really understand what this does other than breaking out of the for loop before it is exhausted.

larsoner: Bumping the loop to 100 or even 1000 seems fine

cbrnr: I understand the point about preferring a for loop over a while loop, but can we be more explicit then? The goal of this part of the test is to reduce the number of visible channels from 20 to 1, so this requires exactly 19 iterations. What's the point of having 100 and breaking out prematurely?

PR #357, opened 2025-08-11
Move test-specific functions to tests

cbrnr:
This keeps the code base clean and also makes the class shorter.

### DIFF
--- a/src/mne_qt_browser/_pg_figure.py
+++ b/src/mne_qt_browser/_pg_figure.py
@@ -41,14 +41,12 @@ from pyqtgraph import (
     setConfigOption,
 )
 from qtpy.QtCore import (
-    QEvent,
     QSettings,
     QSignalBlocker,
     QThread,
     Signal,
 )
-from qtpy.QtGui import QIcon, QMouseEvent
-from qtpy.QtTest import QTest
+from qtpy.QtGui import QIcon
 from qtpy.QtWidgets import (
     QAction,
     QActionGroup,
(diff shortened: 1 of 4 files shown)


cbrnr:
> Is the idea to copy the _fake_click that currently lives here to live in MNE-Python instead?

Not necessarily. I'd prefer if they still lived in this package, just not in the main `MNEQtBrowser` class. Ideally, these methods should only be available to the tests. I've already done that for the tests in this package, but MNE-Python tests also need them.

larsoner:
Okay with me to have them live in `tests/utils.py` and use `try/except` at the MNE-Python end to prefer that pattern

larsoner:
Oh sorry I hadn't looked that closely, didn't realize they were part of the abstract base class. I'd leave them in that case. Seems like more work than it's worth to move them around / out of the class. And it is kind of nice having both backends implement the same method, even if it is only used in tests. IIRC that was part of design decisions made by @drammock so he should probably weigh in too.

cbrnr:
I would strongly argue for separating the actual functionality from test functionality.

cbrnr:
I think it's a bad pattern to mix in test methods. This is not the purpose of that class. I don't like injecting either, I just did that because it was the least invasive solution. We could also subclass it and add the test methods that way.

drammock:
My general feeling aligns with @larsoner:

> Seems like more work than it's worth to move them around / out of the class.

However: if we can get it to work cleanly, I am OK with eliminating those methods in the main class definition. I think maybe something like this might work:

1. define a function to inject the test-only methods (done)
2. make sure that function is API-available so MNE-Python tests can use it
3. add a fixture to `conftest.py` that monkey-patches the class to inject the test-only methods
4. mark that fixture as `auto-use` so we don't have to write `add_test_browser_methods` in every single test where it's needed

PR #358, opened 2025-08-11
Add local test_raw.fif

cbrnr:
This makes it possible to run the entire test suite without having to clone and `pip install -e` the MNE-Python repo.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ exclude = [
   "/*.yml",
   "/screenshot.png",
   "/CONTRIBUTING.md",
+  "/tests",
 ] # tracked by git, but we don't want to ship those files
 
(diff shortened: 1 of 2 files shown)


larsoner:
Do we need to do anything to `pyproject.toml` to make sure we exclude it from the release? Can you check that the built wheel doesn't contain it? Don't want to increase the distributed wheel (or sdist) by 13MB

cbrnr:
Good point, since we just moved the tests outside of src I'm pretty sure that it is not included in either the src distribution or the wheel. But I'll check tomorrow (have to build locally since we don't seem to have these as build artifacts).

cbrnr:
@larsoner feel free to merge if you're happy!

drammock:
thanks @cbrnr

PR #400, opened 2026-03-11
Warn about antialiasing without OpenGL on Linux

### LINKED ISSUE #355: Some operations are extremely slow on Wayland
### Description of the problem

Some operations like dragging the rectangle in the overview bar or dragging a scrollbar indicator are extremely slow on Linux. Everything else behaves normally, for example clicking anywhere in the overview bar or in the scrollbar is instantaneous. Navigating with the arrow keys (up/down, left/right) is also snappy.

### Steps to reproduce

Open the browser and drag the rectangle in the overview bar. Or drag a scrollbar indicator.

### Expected results

There shouldn't be any performance issues. I've tried this on macOS, and it works normally.

### Actual results

On my Linux system (using Wayland), these operations are extremely slow, to the point that I get a popup dialog with "python3 is not responding".

### Additional information

Installing `pyopengl` has no effect.

```
Platform             Linux-6.15.9-arch1-1-x86_64-with-glibc2.42
Python               3.13.3 (main, Apr  9 2025, 04:03:52) [Clang 20.1.0 ]
Executable           /home/clemens/Projects/mne-qt-browser/.venv/bin/python
CPU                  AMD Ryzen 7 5800X 8-Core Processor (16 cores)
Memory               15.5 GiB

Core
├☑ mne               1.10.0 (latest release)
├☑ numpy             2.3.2 (unknown linalg bindings (threadpoolctl module not found: No module named 'threadpoolctl'))
├☑ scipy             1.16.1
└☑ matplotlib        3.10.5 (backend=qtagg)

Numerical (optional)
└☐ unavailable       sklearn, numba, nibabel, nilearn, dipy, openmeeg, cupy, pandas, h5io, h5py

Visualization (optional)
Fontconfig warning: using without calling FcInit()
├☑ qtpy              2.4.3 (PySide6=6.9.0)
├☑ pyqtgraph         0.13.7
├☑ mne-qt-browser    0.8.0.dev32+g62c9fe1
└☐ unavailable       pyvista, pyvistaqt, vtk, ipympl, ipywidgets, trame_client, trame_server, trame_vtk, trame_vuetify

Ecosystem (optional)
└☐ unavailable       mne-bids, mne-nirs, mne-features, mne-connectivity, mne-icalabel, mne-bids-pipeline, neo, eeglabio, edfio, mffpy, pybv
```

cbrnr:
Performance on Linux without OpenGL is extremely poor. In contrast to what I mentioned in #355, installing PyOpenGL actually improves performance a lot, so I think we should at least show an appropriate warning. I've also removed PyOpenGL as a requirement on macOS, since this package has no effect on that platform. So I would say this fixes #355.

Example:

```python
import mne

raw = mne.io.read_raw_fif(
    sample.data_path() / "MEG" / "sample" / "sample_audvis_raw.fif"
)
raw.plot(splash=False)
```

The last line needs to be changed as follows to achieve a decent performance:

```
raw.plot(splash=False, use_opengl=True)
```

Note that it would be kind of nice to expose an `antialiasing` parameter (similar to `use_opengl`), but for that we'd have to change `BaseRaw.plot` first.

Closes #375

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,6 @@ dependencies = [
   "scooby",
   "mne>=1.0",
   "pyqtgraph>=0.12.3",
-  "pyopengl; platform_system=='Linux'",
   "packaging",
   "darkdetect",
   "qdarkstyle",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
   "scooby",
   "mne>=1.0",
   "pyqtgraph>=0.12.3",
-  "pyopengl; platform_system=='Darwin'",
+  "pyopengl; platform_system=='Linux'",
   "packaging",
   "darkdetect",
   "qdarkstyle",
--- a/src/mne_qt_browser/_pg_figure.py
+++ b/src/mne_qt_browser/_pg_figure.py
@@ -496,6 +496,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):  # type: i
             platform.system() == "Linux"
             and self.mne.antialiasing
             and not self.mne.use_opengl
+            and not check_version("pyqtgraph", "0.14")
         ):  # pragma: no cover
             warn(
                 "On Linux, antialiasing without OpenGL can result in poor performance. "
--- a/src/mne_qt_browser/_pg_figure.py
+++ b/src/mne_qt_browser/_pg_figure.py
@@ -460,8 +460,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):  # type: i
         self._add_scalebars()
 
         # Check for OpenGL
-        # If a user doesn't specify whether or not to use it, disable it (PyOpenGL is an
-        # optional requirement)
+        # If a user doesn't specify whether or not to use it, disable it
         opengl_key = "MNE_BROWSER_USE_OPENGL"
         if self.mne.use_opengl is None:  # default: opt-in
             config_val = get_config(opengl_key, "").lower()
--- a/src/mne_qt_browser/_pg_figure.py
+++ b/src/mne_qt_browser/_pg_figure.py
@@ -467,7 +467,7 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):  # type: i
             config_val = get_config(opengl_key, "").lower()
             self.mne.use_opengl = config_val == "true"
 
-        if self.mne.use_opengl:
+        if self.mne.use_opengl and not check_version("pyqtgraph", "0.14"):
             try:
                 import OpenGL
             except (ModuleNotFoundError, ImportError):
--- a/src/mne_qt_browser/_pg_figure.py
+++ b/src/mne_qt_browser/_pg_figure.py
@@ -222,7 +222,7 @@ class _PGMetaClass(type(QMainWindow), type(BrowserBase)):
     pass
 
 
-class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
+class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):  # type: ignore
     """A PyQtGraph-backend for 2D data browsing."""
 
     gotClosed = Signal()
@@ -460,8 +460,8 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
         self._add_scalebars()
 
         # Check for OpenGL
-        # If a user doesn't specify whether or not to use it, disable it (performance
-        # differences seem minimal, and PyOpenGL is an optional requirement)
+        # If a user doesn't specify whether or not to use it, disable it (PyOpenGL is an
+        # optional requirement)
         opengl_key = "MNE_BROWSER_USE_OPENGL"
         if self.mne.use_opengl is None:  # default: opt-in
             config_val = get_config(opengl_key, "").lower()
@@ -472,9 +472,9 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
                 import OpenGL
             except (ModuleNotFoundError, ImportError):
                 warn(
-                    "PyOpenGL was not found and OpenGL cannot be used. "
-                    "Consider installing pyopengl with pip or conda or set "
-                    '"use_opengl=False" to avoid this warning.'
+                    "PyOpenGL was not found so OpenGL cannot be used. Consider "
+                    "installing PyOpenGL or setting 'use_opengl=False' to avoid this "
+                    "warning."
                 )
                 self.mne.use_opengl = False
             else:
@@ -487,10 +487,20 @@ class MNEQtBrowser(BrowserBase, QMainWindow, metaclass=_PGMetaClass):
             and not check_version("pyqtgraph", "0.13.8")
         ):  # pragma: no cover
             warn(
-                "On macOS, Qt 6.10.0, pyqtgraph < 0.13.8, with use_opengl=True results "
-                f"in very slow performance. Consider downgrading {API_NAME}, "
-                "setting use_opengl=False (which can hurt performance), or "
-                "upgrading pyqtgraph once 0.13.8 is released"
+                "On macOS, the combination of Qt 6.10.0 and pyqtgraph < 0.13.8 with "
+                "'use_opengl=True' results in poor performance. Consider downgrading "
+                f"{API_NAME}, setting 'use_opengl=False' (which can hurt performance), "
+                "or upgrading pyqtgraph."
+            )
+        if (
+            platform.system() == "Linux"
+            and self.mne.antialiasing
+            and not self.mne.use_opengl
+        ):  # pragma: no cover
+            warn(
+                "On Linux, antialiasing without OpenGL can result in poor performance. "
+                "Consider installing PyOpenGL or disabling antialiasing in the browser "
+                "settings (or by pressing 'l')."
             )
         # Initialize BrowserView (inherits QGraphicsView)
         self.mne.view = BrowserView(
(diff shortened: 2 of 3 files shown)


larsoner:
> We should probably also include the recommendation to set the Qt rendering platform to X11 via export QT_QPA_PLATFORM=xcb

This makes sense to put somewhere, not sure where, though

> In fact, X11 is still so much faster, but only if PyOpenGL is available and enabled

On my Linux machine, both with and without PyOpenGL things are fast. This is really going to be machine/hardware dependent (graphics drivers, graphics card, CPU, etc.) -- I could imagine on some systems at least OpenGL might actually be worse (or at least there was a time I would have expected it for AMD graphics drivers, might still be true for software rendering...). But if we expect it to be better for most (and we probably do?), then we could update `use_opengl=None` to mean "True on Linux if PyOpenGL is installed, False otherwise" rather than "False" as it currently does (ignoring for now the ability to configure it, which we should still respect).

pijyoi:
By the way, since pyqtgraph 0.14.0, OpenGL rendering of 2D graphics does **not** make use of PyOpenGL. Instead, if the user enables OpenGL (with config option `useOpenGL=True`), it uses Qt's QtOpenGL module. So MNE should not be setting `useOpenGL=True` based on the presence of PyOpenGL.

For rendering of 3D graphics, pyqtgraph still requires PyOpenGL.

larsoner:
We only use for 2D so yeah we should update our message !

cbrnr:
Yep, I can confirm that this has no effect with pyqtgraph >= 0.14 anymore. I'll adapt the message!

cbrnr:
If we simply require pyqtgraph ≥ 0.14 we can remove the pyopengl requirement entirely, right? We're only concerned with 2D so I guess this would be the best way to proceed.

larsoner:
Looks like that came out only about 5 months ago, so I think it's still too new (would be best to give people a two year window if possible). I don't think it complicates things too much for us to continue supporting older versions

cbrnr:
OK, just adapt the message then? I wouldn't keep it around that long though, why would someone not be able to upgrade to 0.14? And even if this is really not possible, they could stick with the current mne-qt-browser version.

larsoner:
It's a variant of https://scientific-python.org/specs/spec-0000/ that we've discussed and adopted in MNE-Python (most recently maybe [here](https://github.com/mne-tools/mne-python/issues/13449#issuecomment-3412146757) but it tends to be what we've tried to do over many years now for good reasons), useful to follow the same pattern here if it's not too onerous to do so (I think mne-qt-browser can be a bit more flexible than MNE-Python but here it's just a matter of tweaking a message)

cbrnr:
Yeah, I still disagree, but I won't argue and instead just change the message.

cbrnr:
@larsoner this should be good to go.

tests/conftest.py
```diff
@@ -13,6 +14,14 @@
...
 def pytest_configure(config):
     """Configure pytest options."""
+    # Enable OpenGL on Linux
+    if platform.system() == "Linux" and "MNE_BROWSER_USE_OPENGL" not in os.environ:
+        try:
+            import OpenGL  # noqa: F401
+
+            os.environ["MNE_BROWSER_USE_OPENGL"] = "true"
+        except ImportError:
+            pass
```
larsoner: I'd rather not force this on everyone running tests. You could add the env var to the CI config if you want them to have antialiasing enabled though

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
   "scooby",
   "mne>=1.0",
   "pyqtgraph>=0.12.3",
-  "pyopengl; platform_system=='Darwin'",
+  "pyopengl; platform_system=='Linux'",
```
larsoner: Because of

> since pyqtgraph 0.14.0, OpenGL rendering of 2D graphics does not make use of PyOpenGL

I'd rather eliminate this entirely:
```suggestion
```

src/mne_qt_browser/_pg_figure.py
```diff
@@ -472,9 +472,9 @@ def __init__(self, **kwargs):
...
             except (ModuleNotFoundError, ImportError):
                 warn(
-                    "PyOpenGL was not found and OpenGL cannot be used. "
-                    "Consider installing pyopengl with pip or conda or set "
-                    '"use_opengl=False" to avoid this warning.'
+                    "PyOpenGL was not found so OpenGL cannot be used. Consider "
+                    "installing PyOpenGL or setting 'use_opengl=False' to avoid this "
+                    "warning."
                 )
                 self.mne.use_opengl = False
```
larsoner, line 478: I think this whole block now needs to be nested in a `if not check_version("pyqtgraph", "0.14"):`

src/mne_qt_browser/_pg_figure.py
larsoner: We should only need to check for PyOpenGL on older versions of pyqtgraph so I think we need to change this to:
```suggestion
        if self.mne.use_opengl and not check_version("pyqtgraph", "0.14"):
```

tests/conftest.py
cbrnr: We can remove it since it isn't necessary with PyQtGraph >= 0.14 anymore. The idea was not to force anything on everyone, because this snippet just sets the environment variable if `OpenGL` is installed.

pyproject.toml
cbrnr: Yes, but weren't you against bumping `pyqtgraph` to `>=0.14`? I think this line is still needed for Linux in the case someone is still using `<0.14`.

src/mne_qt_browser/_pg_figure.py
cbrnr: Agreed. But should we not also deprecate the `use_opengl` parameter as well then?

tests/conftest.py
larsoner: > this snippet just sets the environment variable if OpenGL is installed.

By setting this env var, it would silently cause people who have PyOpenGL installed to always use OpenGL in these tests. By `get_config` precedence, an env var overrides MNE-Python config file variables and the MNE default behavior when not set (`False`). This is what I mean by forcing the behavior on people, does that make sense?

pyproject.toml
larsoner: My understanding is that on macOS it was really required for things to function at all previously. On Linux, things work, but for some people in some configurations it's noticeably slower. I don't think it rises to an absolute (non-optional) requirement for Linux, even for PyQtGraph < 0.14

src/mne_qt_browser/_pg_figure.py
larsoner: No I think it's useful to have OpenGL use be optional still. Historically OpenGL support has been bad on some drivers on some systems, especially (annoyingly enough) on Linux

At MNE-Python end we could change the default to `True` when PyQtGraph is new enough, though

cbrnr: Wait, now I'm confused. The only reason to keep `use_opengl` around is if we still allow users to have `pyqtgraph < 0.14`, right?

pyproject.toml
cbrnr: I always had issues on Linux and never on macOS. But I'm fine with removing the requirement as it should not be necessary anymore, at least with `pyqtgraph >= 0.14`.

tests/conftest.py
cbrnr: Kind of, you mean if they explicity set `use_opengl=False` this would override it silently? Yeah, that's probably not what we want, although I have always been assuming that this change would be for the better (or at least wouldn't influence anything negatively).

src/mne_qt_browser/_pg_figure.py
larsoner: No I don't think so. There might be users who have configurations where using OpenGL is not beneficial (or even working correctly!). So we want to allow them to say `use_opengl=False` (which gets passed to PyQtGraph as `BrowserView(..., useOpenGL=use_opengl)` and used / goes into effect for all versions of PyQtGraph).

tests/conftest.py
larsoner: > you mean if they explicity set use_opengl=False this would override it silently? 

No not quite. I mean currently (on `main`) for devs on Linux, whether OpenGL is used in tests is:

1. False by default (by MNE default), or
2. Opt-in with a key in `~/.mne/mne-python.json`, or
3. Opt-in with a env var `MNE_BROWSER_USE_OPENGL`

If you keep this change as-is, all Linux devs with PyOpenGL installed at all (which includes anyone who uses our installers) will now have OpenGL silently enabled, overriding the MNE default (False) and their own config JSON (where they might have set it to False), i.e., is a behavior change for anyone who hasn't otherwise opted in somehow

src/mne_qt_browser/_pg_figure.py
cbrnr: But why? I thought we wanted to get rid of it, so it will always be `False`.

larsoner: Okay I think I understand the (a?) source of confusion. It sounds like you think `use_opengl` means "use PyOpenGL Python library". I have been thinking it means "use OpenGL for rendering".

Based on :

> since pyqtgraph 0.14.0, OpenGL rendering of 2D graphics does not make use of PyOpenGL. Instead, if the user enables OpenGL (with config option useOpenGL=True), it uses Qt's QtOpenGL module.

I think in the context of pyqtgraph, `useOpenGL=True | False` is still supported, it's just that OpenGL calls get routed through Qt instead of needing PyOpenGL when `useOpenGL=True`. I think we should think of it the same way, where `use_opengl` really means "use OpenGL for rendering" (and whether or not PyOpenGL is involved at all just has to do with the pyqtgraph version and whether or not `use_opengl` is effectively True or False).

cbrnr: Ah! This makes sense! Thanks!

cbrnr:
@larsoner I think this should be good to go.

larsoner (approved):
Marking for merge when green, thanks @cbrnr !

Issue #408, opened 2026-04-24
BUG: epoch_colors desynchronize when scrolling in EEG plots in default backend


DavoudYneuro:
### Description of the problem

When using the `epoch_colors` parameter in MNE's `epochs.plot()` function, the custom colors display correctly on the initial view but become desynchronized when scrolling, resizing, or interacting with the plot. Instead of showing properly colored EEG traces, the plot shows vertical lines or color boxes overlaid on the data.

## Affected Versions

- **MNE-Python**: All versions with epoch_colors support
- **mne-qt-browser**: All versions (default backend)
- **Backends affected**: Both Qt (mne-qt-browser) and Matplotlib

### Steps to reproduce

```Python
import mne
import numpy as np

# Create sample epochs
info = mne.create_info(ch_names=['Fz', 'Cz', 'Pz', 'Oz'], sfreq=256., ch_types='eeg')
n_epochs = 50
n_channels = 4
n_times = 256
data = np.random.randn(n_epochs, n_channels, n_times) * 1e-6

epochs = mne.EpochsArray(data, info)

# Create custom epoch_colors with distinct patterns
# This makes it easy to see if colors are misaligned
epoch_colors = []
for i in range(n_epochs):
    if i % 5 == 0:
        colors = ['red', 'blue', 'red', 'red']  # Every 5th epoch is red
    elif i % 3 == 0:
        colors = ['blue', 'blue', 'red', 'blue']  # Every 3rd epoch is blue
    elif i % 2 == 0:
        colors = ['green', 'black', 'green', 'green']  # Even epochs are green
    else:
        colors = ['black', 'black', 'green', 'black']  # Odd epochs are black
    epoch_colors.append(colors)

# Plot with custom epoch_colors
fig = epochs.plot(epoch_colors=epoch_colors, n_epochs=10, block=True)


# STEPS TO REPRODUCE:
# 1. Initial view shows epochs ✓ CORRECT
# 2. Press Page Down or scroll right
# 3. View now shows other epochs
# 4. BUG: Colors show as vertical lines or are at wrong positions ✗ BROKEN
```

### Expected results

Custom colors assigned to specific epochs via the `epoch_colors` parameter should:
- Display correctly on initial render
- Remain synchronized with their assigned epochs when scrolling
- Maintain correct colors when resizing the window
- Show colored EEG traces (not vertical lines or boxes)

[...]

PR #410, opened 2026-05-01
FIX: MNE_BROWSER_PRECOMPUTE=true epoch rejection creating solid blocks

DerAndereJohannes:
When clicking an epoch to mark/unmark it as bad while precompute is enabled, the epoch column turns into a solid colored block instead of showing the waveform highlighted. The included epochs also appear as solid blue. Scrolling away and back with the arrow keys temporarily restores included epochs but the rejected epoch remains a solid red bar.

There also seemed to be issues when changing the colours on the different channels such as in the first mentioned issue below.

#### Reference issue

- https://github.com/mne-tools/mne-qt-browser/issues/408
  - Note that I just looked at the qt backend and did not check the matplotlib backend
- https://github.com/mne-tools/mne-qt-browser/issues/270

The fix is based in the `update_data()` function where the previous version would directly divide the data itself in place through a numpy view. The fix moves the scaling out of the data entirely and into the Qt transform.

For the epoch fix, I also had to add a redraw at the end of `_precompute_finished()`. Epoch rejection calls `update_data()` directly without going through `_update_data()` first, so without the explicit redraw, `self.mne.data` was never transitioned to the precomputed buffer before epoch rejection tried to use it.

#### Additional information

I have tried the fix on MacOS and Linux systems. Let me know if I should change anything else and please do let me know if this works for you too.

Image of it working on epochs :-)

image

### DIFF
--- a/src/mne_qt_browser/_graphic_items.py
+++ b/src/mne_qt_browser/_graphic_items.py
@@ -527,7 +527,9 @@ class DataTrace(PlotCurveItem):
     def update_scale(self):  # noqa: D102
         transform = QTransform()
         if self.mne.data_precomputed:
-            transform.scale(1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type])
+            transform.scale(
+                1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type]
+            )
         else:
             transform.scale(1.0, self.mne.scale_factor)
         self.setTransform(transform)
@@ -566,7 +568,9 @@ class DataTrace(PlotCurveItem):
 
         transform = QTransform()
         if self.mne.data_precomputed:
-            transform.scale(1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type])
+            transform.scale(
+                1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type]
+            )
             data = self.mne.data[self.order_idx]
         else:
             transform.scale(1.0, self.mne.scale_factor)
--- a/src/mne_qt_browser/_graphic_items.py
+++ b/src/mne_qt_browser/_graphic_items.py
@@ -526,7 +526,10 @@ class DataTrace(PlotCurveItem):
     @propagate_to_children
     def update_scale(self):  # noqa: D102
         transform = QTransform()
-        transform.scale(1.0, self.mne.scale_factor)
+        if self.mne.data_precomputed:
+            transform.scale(1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type])
+        else:
+            transform.scale(1.0, self.mne.scale_factor)
         self.setTransform(transform)
 
         if self.mne.clipping is not None:
@@ -561,11 +564,14 @@ class DataTrace(PlotCurveItem):
             connect = "all"
             skip = True
 
+        transform = QTransform()
         if self.mne.data_precomputed:
+            transform.scale(1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type])
             data = self.mne.data[self.order_idx]
-            data /= self.mne.scalings[self.ch_type]
         else:
+            transform.scale(1.0, self.mne.scale_factor)
             data = self.mne.data[self.range_idx]
+        self.setTransform(transform)
         times = self.mne.times
 
         # Get decim-specific time if enabled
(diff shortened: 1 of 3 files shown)


src/mne_qt_browser/_graphic_items.py
marsipu, line 542:
```suggestion
```

marsipu:
```suggestion
        # update_data contains update_scale
        if self.mne.clipping is None:
            self.update_data()
        else:
            self.update_scale()
```

src/mne_qt_browser/_graphic_items.py
```diff
@@ -561,11 +566,16 @@ def update_data(self):
...
         if self.mne.data_precomputed:
+            transform.scale(
+                1.0, self.mne.scale_factor / self.mne.scalings[self.ch_type]
+            )
             data = self.mne.data[self.order_idx]
-            data /= self.mne.scalings[self.ch_type]
         else:
+            transform.scale(1.0, self.mne.scale_factor)
             data = self.mne.data[self.range_idx]
+        self.setTransform(transform)
```
marsipu:
```suggestion
        # Apply scale transformation
        self.update_scale()
        
        # Get data
        if self.mne.data_precomputed:
            data = self.mne.data[self.order_idx]
        else:
            data = self.mne.data[self.range_idx]
```

marsipu (changes requested):
Thanks a lot Johannes for this nice fix. On Windows, your fix works well too. Also, the unit-scaling by dialog from #278 still seems to work.
The transformation is appearing now twice in the code, in `DataTrace.update_scale` and `DataTrace.update_data`. imho it would be better to have this only once to avoid confusion and future bugs of the same sort. I noticed, that `update_scale` is only used by `DataTrace.__init__` and `MNEQtBrowser.scale_all`. Both already consider the case when clipping is set, so that seems also duplicated. Thus for further improvement also unrelated to your changes I would suggest the following changes.
The only disadvantage could be reduced performance since `update_scale` would be called now on every `update_data`. But I would expect applying the transformation to be a very cheap operation.

DavoudYneuro:
@DerAndereJohannes Thank you for the fix!

larsoner:
@DerAndereJohannes let me know if you need help addressing the review comments, happy to push some commits if it helps

DerAndereJohannes:
Hey, sorry for the late response and thank you for the confirmations that it is working.                                                                                
                                                                                                                                                                          
I reviewed your edits and thought about a way of getting around the *small* performance disadvantage. I thought of just putting the transform into its own method and just letting `update_scale` and `update_data` call it. Since it does not have the propagate decorator like calling `self.update_scale()` directly. 

I hope that the change is suitable, if not then I would also be happy to revert and use your changes.

Thank you for your patience and your comments in advance.

marsipu (approved):

marsipu:
@DerAndereJohannes Thanks a lot, I think that is even better. I just compared the benchmarks of the CIs and the ones from this PR (right side) is not slower, even a bit faster (although I can't explain why it is even faster for precompute=False, where you didn't really change anything). But let's take it.
grafik

marsipu:
@larsoner I had to add PySide6=6.11.1 to the exclusion string for tests, since it produced a lot of Ci-failures. I see that this is not the first version we exclude. Should this version be also excluded on the mne-side? What is our general policy there?
The errors seem to be again on the level of the c++-binding with "fatal access violation" and "segmentation faults". I could at least reproduce some of them locally on Windows with 6.11.1. but usually I find these quite hard to solve.

DerAndereJohannes:
I think you cannot rely on github actions for performance benchmarks since I am sure that the compute resources are shared. I guess Sunday is a good time to run them ;-)

larsoner:
> @larsoner I had to add PySide6=6.11.1 to the exclusion string for tests, since it produced a lot of Ci-failures. I see that this is not the first version we exclude. Should this version be also excluded on the mne-side? What is our general policy there?

Yeah if we exclude here we should exclude there, too. I'll work on it in https://github.com/mne-tools/mne-qt-browser/pull/411 and https://github.com/mne-tools/mne-python/pull/13899 to see if I can get it to work.

In the meantime we can merge this I think and I can remove the pin in https://github.com/mne-tools/mne-qt-browser/pull/411

larsoner:
Thanks @DerAndereJohannes !

PR #423, opened 2026-07-03
Add option to show zero line

### LINKED ISSUE #353: Add option to show zero for each channel
### Describe the problem

Currently, it is rather difficult to determine if a given data point lies below or above zero, which can be important e.g. when working with EEG (ERPs). Each channel has a tick on the left y-axis indicating zero, but this only helps when the data point of interest is very close to this indicator.

Image

### Describe your solution

A continuous horizontal line (in light gray) would make that much easier. Since the zero point is not important for all kinds of signals, this should definitely be an option to toggle on and off.

### Describe possible alternatives

I don't think there's any alternative.

### Additional context

_No response_

cbrnr:
Companion PR to https://github.com/mne-tools/mne-python/pull/14018 (adds a zero line and fixes #353). This requires that the MNE-Python PR gets merged and is available for the tests to pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ tests = [
   "pytest-harvest",
   "pytest-error-for-skips",
   "sphinx-gallery",
+  "refleak >=0.2.1",
 ]
 
 [project.urls]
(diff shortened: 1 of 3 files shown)


larsoner:
> This requires that the MNE-Python PR gets merged and is available for the tests to pass.

We don't want these tests to fail on old versions of MNE in tests or for users, so the code should use a `getattr(self.mne, "zero_line_visible", False)` or similar

cbrnr (commented):
@larsoner the behavior matches the one from the MPL backend, this should be good to go if tests come back green.

larsoner (approved):
Failures are unrelated but I've pushed a commit that should fix them. Marking for merge-when-green, thanks in advance @cbrnr !

PR #424, opened 2026-07-03
Fix leaky tests and show DS warning

cbrnr:
I noticed that after running the test suite, the downsampling factor was persisted to 17 in my settings. This PR fixes this by using a temporary settings file.

In addition, I've added a statusbar warning when the downsampling factor is much higher than what "auto" would use. If it's more than 3x higher, the message is shown in red.


larsoner (approved):
Nice, I was wondering why my raw plots were all messed up! 🤦

PR #425, opened 2026-07-09
FIX: Checks

larsoner:
On Pyside6 6.11.1 I sometimes get when closing windows:
```
/home/larsoner/python/mne-qt-browser/src/mne_qt_browser/_utils.py:36: RuntimeWarning: libpyside: Failed to disconnect (None) from signal "triggered()".
  sig.disconnect()
/home/larsoner/python/mne-qt-browser/src/mne_qt_browser/_utils.py:36: RuntimeWarning: libpyside: Failed to disconnect (None) from signal "triggered()".
  sig.disconnect()
/home/larsoner/python/mne-qt-browser/src/mne_qt_browser/_utils.py:36: RuntimeWarning: libpyside: Failed to disconnect (None) from signal "triggered()".
  sig.disconnect()
/home/larsoner/python/mne-qt-browser/src/mne_qt_browser/_utils.py:36: RuntimeWarning: libpyside: Failed to disconnect (None) from signal "triggered()".
  sig.disconnect()
/home/larsoner/python/mne-qt-browser/src/mne_qt_browser/_utils.py:36: RuntimeWarning: libpyside: Failed to disconnect (None) from signal "triggered()".
  sig.disconnect()
```
This helps swallow those. I also intermittently got some `AttributeError`s so this safeguards a `delattr`

PR #428, opened 2026-07-17
Fix multiple bugs and inefficiencies

larsoner:
Looking at the code with Claude Fable 5 there were a number of small issues that were easily fixed. I also folded in my changes from #425

Closes #425


larsoner:
@drammock this is another one that fixes a bunch of stuff at once... happy to split out into multiple PRs if you want to review and it would help

larsoner:
Windows failure should be fixed by https://github.com/mne-tools/mne-python/pull/14065

larsoner:
Will split into separate PRs

PR #429, opened 2026-07-18
Fix suppression of prefixed PySide disconnect warnings

mh105:
#### Reference issue

N/A

PySide now prefixes its no-connections warning with `libpyside:`, so the existing `warnings.filterwarnings` message pattern no longer matches it. Closing the browser consequently emits repeated `RuntimeWarning: libpyside: Failed to disconnect ...` messages.

This updates `_disconnect` to accept the optional `libpyside: ` prefix while retaining the original message match. A regression test covers the prefixed warning and verifies that unrelated runtime warnings are not suppressed.

#### Additional information

Validated with:

- `pytest -q -o addopts='' tests/test_pg_specific.py::test_disconnect_warning_filter`
- `ruff check src/mne_qt_browser/_utils.py tests/test_pg_specific.py`
- `ruff format --check src/mne_qt_browser/_utils.py tests/test_pg_specific.py`
- An end-to-end `Raw.plot()` Qt-browser close using PySide6 6.11.1, which emitted no disconnect runtime warnings after the change.

### DIFF
--- a/src/mne_qt_browser/_utils.py
+++ b/src/mne_qt_browser/_utils.py
@@ -31,7 +31,9 @@ def _disconnect(sig, *, allow_error=False):
     try:
         with warnings.catch_warnings():
             warnings.filterwarnings(
-                "ignore", "Failed to disconnect", category=RuntimeWarning
+                "ignore",
+                "(?:libpyside: )?Failed to disconnect",
+                category=RuntimeWarning,
             )
             sig.disconnect()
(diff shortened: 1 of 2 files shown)


larsoner (approved):
I also just hit this, if it's good to go can you take it out of draft mode and I'll merge? Thanks!

mh105:
Just opened it for review! Yeah first time PRing into this repo, so I wanted to observe what CIs are involved first.

PR #430, opened 2026-07-18
Fix CIs and simplify

larsoner:
We don't need to triage `opengl` installs anymore since PyQtGraph doesn't use `pyopengl`. Simplify CIs. Deduplicate blocks. Fix `maint/*` branch (it wasn't using it!).


larsoner:
This one seems like a straightforward CI fix PR (doesn't change anything about what we actually test... other than fixing it to test what we think we actually test currently!) so I'll adjust CI requirements once it's ready and self-mark for squash+merge-when-green

PR #431, opened 2026-07-18
MAINT: Isolate MNE config and shared fixtures

larsoner:
Avoid any possibility of messing with MNE config with a session-scoped fixture, and avoid mutating data from a shared fixture

Found and implemented with Claude Fable 5, and I understand and vouch for all changes


larsoner:
I'm looking into some visual tweaks and got bit by this again so I'll merge (it's pretty uncontroversial I think!)

PR #434, opened 2026-07-18
FIX: Bug fixes and cleanups in _widgets

larsoner:
Small things:

- Annotation description combobox: rebuilding it clears then repopulates, emitting currentIndexChanged(-1)/(0) which clobbered `mne.current_description`, so block signals during the rebuild
- `bad_epoch_rect_dict` is keyed by epoch number on `main`, which only equals the boundary-times index when no epochs were dropped, so map number to index via `inst.selection` before repositioning
- `setPageStep` in scrollbar units (duration * step_factor) so the slider length reflects the visible fraction
- set-based membership of `ChannelAxis` when pruning ch_texts
- build an onset to index map for `OverviewBar` once instead of scanning annotations.onset per rect on every horizontal scroll

Found and implemented with Claude Fable 5, and I understand and vouch for all changes

### DIFF
--- a/src/mne_qt_browser/_widgets.py
+++ b/src/mne_qt_browser/_widgets.py
@@ -175,8 +175,8 @@ class AnnotationDock(QDockWidget):
         # Update related widgets
         self.weakmain()._setup_annotation_colors()
         self._update_regions_colors()
-        self._update_description_cmbx()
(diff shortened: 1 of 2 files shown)


drammock (approved):
LGTM, but has merge conflicts

PR #442, opened 2026-07-28
FIX: Butterfly mode scale bars and spacing (#276)

### LINKED ISSUE #276: BUG: Butterfly mode scale bars overlap
On latest MNE `main` (so you can use `--butterfly`) you can see:
```
MNE_BROWSER_BACKEND=matplotlib mne browse_raw ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif --butterfly --clipping None
```
| matplotlib | qt |
| -- | -- |
| image |image |

**BUG 1**: If you plot in non-butterfly mode then hit "b" you get a different scale bar (e.g., 800 fT/cm instead of 400 fT/cm):

image

**BUG 2**: I think the spacing might also be off -- there is too much space at the top and bottom of the plot compared to the MPL version (which seems more correct). Some of these differences probably give rise to these differences in the MNE-Python test:

https://github.com/mne-tools/mne-python/pull/12721/files#diff-54ae16f8e47e25048ec09729d5f2175a49995177d97d1c0349bab066cbc90fffR1243-R1244

**BUG 3**: The z order is wrong. Whichever channel type is at the top should have the highest z order probably (it comes earlier in the data order so should be considered higher priority I think).

larsoner:
Closes #276.

Draft until we see this and MNE-Python green. Then:

- [x] switch this back to using MNE `main`
- [x] switch the MNE-Python PR back to using mne-qt-browser `main`
- [x] merge MNE-Python branch (should be green)
- [x] restart CIs here, should come back green
- [x] merge here

Visual confirmation (synthetic data using sawtooth wave with peak-to-peak that should match the scalebars):

| - | main | PR |
| -- | -- | -- |
| mpl | image | image| 
| qt | image | image |

<details>
<summary>MWE</summary>

```
import numpy as np
from scipy.signal import sawtooth

import mne

sfreq = 100.0
period = 50  # samples; 10 whole periods over 5 s keeps the wave exactly zero-mean
n = 500
ph = (np.arange(n) % period) / period
triangle = 1 - 2 * np.abs(2 * ph - 1)
peak_to_peak = dict(grad=400e-13, mag=1000e-15, eeg=20e-6)
n_per_type = 2  # a couple of channels per type, as in any real recording
ch_types, ch_names, data = [], [], []
for ct, pp in peak_to_peak.items():
    for i in range(n_per_type):
        ch_types.append(ct)
        ch_names.append(f"{ct}{i}")
        data.append(pp / 2 * np.roll(triangle, i * period // n_per_type // 2))
info = mne.create_info(ch_names, sfreq, ch_types=ch_types)
raw = mne.io.RawArray(np.array(data), info)
scalings = {ct: pp / 2 for ct, pp in peak_to_peak.items()}
with mne.viz.use_browser_backend("matplotlib"):
    raw.plot(scalings=scalings, clipping=None)
with mne.viz.use_browser_backend("qt"):
    raw.plot(scalings=scalings, clipping=None)
# press "b" in both
```

</details>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -79,7 +79,16 @@ jobs:
         with:
           qt: true
       - run: mne sys_info
-      - run: pytest -m pgtest --cov=mne_qt_browser --cov-report=xml mne-python/mne/report mne-python/mne/viz
+      - run: |
+          # On MNE < 1.13 (i.e. not main) test_plotting_scalebars still asserts the
+          # pre-#276 butterfly scalebar geometry (bars spanning a full y-unit), which
+          # the fix here intentionally changes. The updated test only lives on MNE
+          # main (via the companion PR), so deselect it on the older release branches.
+          deselect=()
+          if [[ "${{ matrix.mne }}" != "main" ]]; then
+            deselect=(--deselect mne-python/mne/viz/tests/test_raw.py::test_plotting_scalebars)
+          fi
+          pytest -m pgtest --cov=mne_qt_browser --cov-report=xml "${deselect[@]}" mne-python/mne/report mne-python/mne/viz
         name: Run MNE-Tests
       - run: pytest tests/test_pg_specific.py --cov-report=xml --cov-append
         name: Run pyqtgraph-specific tests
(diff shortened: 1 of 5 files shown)


drammock (approved):

larsoner:
Hah that was an errant auto-close

PR #449, opened 2026-08-18
Keep Qt windows responsive under marimo

larsoner:
This makes `raw.plot()` start a small Qt event pump when it detects `marimo`, so windows are interactive with no user-side boilerplate, and documents the `marimo` workflow with an example notebook. @drammock locally things work for me on this branch with:
```
$ python -m marimo edit --headless --no-token -p 2718 ~/python/mne-qt-browser/examples/marimo_demo.py
```
and opening `http://localhost:2718`.

I didn't see any differences on macOS or Linux when testing. Having a `mo.ui.run_button` in a separate cell followed by `mo.stop` in a separate cell seems to be a working way to pause until the user somehow indicates that they're done interacting. It's possible some of the non-interactivity you were seeing was due to subsequent steps/cells in the notebook executing -- those will prevent the events from being processed (just like when running with `plt.ion()` in standard Python).

You can also check out https://github.com/mne-tools/mne-python/pull/14179 , and `block=True` should "just work", on this branch (or even if you're not on this branch).

I also drafted https://github.com/marimo-team/marimo/pull/10584 which could fix this at the `marimo` end.

Worked with Claude Opus 5 to make this work on macOS and Linux.


BelizSertcan:
Thanks a lot for working on this. 
We tried this on both Linux and macOS, and it works as intended for Linux.
On Macos, it also works well until the final window. The final winwow does not close cleanly, and gives a spinning beachball and I have to force-quit the Python process. 
I suspect that the event pump stops once the notebook is idle/done, so a window still open at that point is left with no loop to process its close event. This behavior can also be reproduced with your marino test.

larsoner:
Okay I force-pushed a commit that probably fixes that!

examples/marimo_demo.py
```diff
@@ -0,0 +1,41 @@
...
+def _(mo):
+    go = mo.ui.run_button(label="Done inspecting -- continue")
+    go
+    return (go,)
+
+
+@app.cell
+def _(go, mo):
+    # Everything downstream waits here until the button is clicked, and the window stays
+    # interactive the whole time -- this is what block=True should have been.
```
drammock: this comment isn't really accurate (specifically the "downstream" part). `mo.stop(not go.value)` only gates *this cell's* execution; subsequent cells will still run immediately (i.e., not wait for raw.plot() window to be closed) unless the downstream cells are *also* gated with `mo.stop(...)` (or depend on the output of this cell).

```suggestion
    # This cell won't execute until the button is clicked. Assuming
    # no subsequent cells are running, the window should be interactive.
    # There is no way to gate *all* subsequent cells unless you use
    # `raw.plot(..., block=True)`
```

drammock (approved):
I think we're happy with this!  Note the wording change in the comment on the test; Marimo's execution model means we still end up using `block=True` (easier than adding a `mo.stop(...)` to every single subsequent cell).  But this does save us from needing our own function to pump events. Thanks @larsoner. If it's ready from your end, move out of draft mode?

larsoner:
Okay I'll merge this and its companion https://github.com/mne-tools/mne-python/pull/14179 once I add tests there...

PR #450, opened 2026-08-23
Fix annotation labeling

### LINKED ISSUE #210: Annotation labels should not (or not only) appear in the center of the annotation
**Is your feature request related to a problem? Please describe.**
I have some annotations that span several dozens of seconds. When I spot such an annotated segment in my data, there is no simple way for me to figure out the annotation description: I have to cautiously keep scrolling right until I hit the center of the annotation, because the label is centered.

**Describe the solution you'd like**
The label should be anchored to the annotation onset.

**Describe alternatives you've considered**
Another thing I'd love to see is a tooltip that displays the annotation label on hover.

larsoner:
Closes #210

I iterated a bit with Claude Fable 5 (code reviewed and understood by me) to come up with a way to get multiple labels to show up nicely:

1. Center in the visible part of the label
2. Pack horizontally, stacking when needed using a "first-fit greedy interval coloring" algorithm. This uses the fewest number of rows to fit the labels while centering them on their visible portion (this takes about 15 lines which is neat!)

You can see it in action for example with:

```python
import numpy as np
import mne

rng = np.random.default_rng(0)
info = mne.create_info([f"EEG {i:03d}" for i in range(8)], 250, "eeg")
raw = mne.io.RawArray(rng.standard_normal((8, 250 * 60)) * 20e-6, info)
# long, staggered, overlapping annotations of different kinds
for ii, desc in enumerate(("BAD_muscle", "BAD_eog", "rest", "BAD_ACQ_skip", "stim")):
    raw.annotations.append(onset=5 + 4 * ii, duration=30 - 4 * ii, description=desc)
raw.annotations.append(onset=20, duration=0, description="marker")  # zero-duration
raw.plot(duration=8, scalings=dict(eeg=50e-6))  # press → to scroll, hover for tooltip
```

`main`:

Screenshot 2026-08-22 at 17 53 17

PR:

Screenshot 2026-08-28 at 13 19 35

Also seeds this repo with an AGENTS.md/CLAUDE.md that points to MNE-Python's but adds repo-specific stuff, plus adds a better linewidth for HiDPI (don't double it) and a couple minor efficiency improvements.


larsoner:
pip-pre will be red until https://github.com/pyqtgraph/pyqtgraph/pull/3585 lands

larsoner:
@drammock this one is good to go if you want to try it out using the MWE at the top! Or if you're happy enough with the screenshots + my local testing we can merge

PR #452, opened 2026-08-27
ENH: support variable-duration epochs in browser

snesmaeili:
Companion to mne-tools/mne-python#14210, which lets `Epochs` hold trials of
different duration and browses them natively in the Matplotlib backend. This
does the same for PyQtGraph.

A ragged instance refuses `times`, returns a list of per-epoch arrays from
`get_data`, and reports itself through `variable_duration`. MNE-Python passes
the browser `boundary_times` (cumulative real epoch edges), `boundary_samples`,
`n_times` (total real samples across all epochs), and per-epoch `epoch_tmins` /
`epoch_tmaxs`. Nothing is padded.

The arithmetic lives in `_utils.py` as plain functions so it can be tested
without a Qt event loop: `epoch_window` for the seconds a run of epochs spans,
`epoch_index_at` for the epoch under a display position, `latency_at` for that
position relative to the epoch's own event, and `latency_positions` for where a
latency falls in the epochs that reach it. Every ragged path is reached through
`is_variable_duration`, so an older MNE-Python — which has no such attribute —
follows exactly the code it follows today.

**Equal-duration assumptions replaced.** The global time vector multiplied one
epoch length by the epoch count and now spans `n_times`; `xmax` comes from the
last boundary; scrolling and the time-span buttons step whole epochs and re-read
the duration from the boundaries; the crosshair and the vline label convert
through the containing epoch instead of taking a remainder against one duration;
tick spacing follows the shortest epoch; the overview bar maps clicks and the
bad-channel line through the boundaries rather than assuming every epoch
occupies the same width; and the memory estimate counts real samples and
survives a list payload (`_data.nbytes` raises on a list).

A latency exists only in epochs long enough to reach it, so one line is kept per
visible epoch and hidden where the epoch is too short, which also leaves the
count stable while one is dragged. Their bounds now come from the epoch index
rather than being recovered from the position, which picked the previous epoch
whenever a line sat exactly on a boundary.

**Tests.** Five pure-helper tests need no Qt and always run; seven GUI tests plus
a fixed-duration regression skip on an MNE-Python that cannot build a ragged
object, or that does not yet route one to this backend. Locally: 38 passed and 7
skipped against released MNE 1.11, and 45 passed against the #14210 branch with
its backend guard lifted.

**Note on ordering.** #14210 currently raises `NotImplementedError` for any
non-Matplotlib backend, so this cannot be exercised through `epochs.plot()`
until that guard is relaxed to a version check against a released
mne-qt-browser. That follow-up belongs in MNE-Python, not here.

AI assistance: I designed the approach and ran the analyses it is validated
against; Claude Opus 5 wrote the implementation and the tests under my
direction, which I reviewed and tested.


larsoner:
Can you(r Claude) look at the pending guidance in https://github.com/mne-tools/mne-qt-browser/pull/450/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9 and make any adjustments necessary based on those contents (as if it had read them from the repo root)?

PR #458, opened 2026-09-22
Fix macOS failure

larsoner:
Let's see if we can isolate the Qt6 crash


larsoner:
Being fixed in pyqtgraph in https://github.com/pyqtgraph/pyqtgraph/pull/3640

larsoner:
Upstream opened as https://codereview.qt-project.org/c/qt/qtbase/+/773358

